PR #9, opened 2019-06-24
Makes BigInteger print hex instead of wrong decimal

kobigurk:
Currently, the BigInteger print iterates over the u64 comprising the BigInteger, and prints them in decimal concatenated, left padded with zeros to 20 characters.

This pull request modifies this to a more useful representation in hexadecimal.


Pratyush:
Thanks for the PR! Merged.

PR #10, opened 2019-06-24
Fixes comments for BLS12-377 G2 generators

kobigurk:
Currently, the comments for the G2 generators for BLS12-377 don't match the actual numbers.


Pratyush:
Thanks for the PR! I only made some changes to fix a conflict from another PR.

PR #23, opened 2019-08-16
edit Travis config for CI to typecheck tests as well

savil:
Some questions:

1. Would this make sense to do?
2. Should one also do `cargo test` so that no change gets added which breaks tests?


savil:
hmm not actually sure how to test this except waiting for Travis to report results :-)

savil:
oh oops, just saw #22

PR #36, opened 2019-10-17
Adds Bowe-Hopwood hash

kobigurk:
* Adds 3-bit lookup with conditional negation in Groups
* Adds Bowe-Hopwood implementation according to the Zcash spec
* Adds add_bool_as_coeff and three_bit_cond_neg in Fields

### DIFF
--- a/algebra/src/curves/edwards_bls12/mod.rs
+++ b/algebra/src/curves/edwards_bls12/mod.rs
@@ -2,7 +2,7 @@ use crate::field_new;
 use crate::{
     biginteger::BigInteger256,
     curves::{
-        models::{ModelParameters, TEModelParameters},
+        models::{ModelParameters, TEModelParameters, MontgomeryModelParameters},
         twisted_edwards_extended::{GroupAffine, GroupProjective},
     },
     fields::edwards_bls12::{fq::Fq, fr::Fr},
@@ -55,12 +55,33 @@ impl TEModelParameters for EdwardsParameters {
     /// Generated randomly
     const AFFINE_GENERATOR_COEFFS: (Self::BaseField, Self::BaseField) = (GENERATOR_X, GENERATOR_Y);
 
+    type MontgomeryModelParameters = EdwardsParameters;
+
     /// Multiplication by `a` is just negation.
     /// Is `a` 1 or -1?
     #[inline(always)]
     fn mul_by_a(elem: &Self::BaseField) -> Self::BaseField {
         -*elem
     }
+}
+
+impl MontgomeryModelParameters for EdwardsParameters {
+    /// COEFF_A = 0x8D26E3FADA9010A26949031ECE3971B93952AD84D4753DDEDB748DA37E8F552
+    const COEFF_A: Fq = field_new!(Fq, BigInteger256([
+        13800168384327121454u64,
+        6841573379969807446u64,
+        12529593083398462246u64,
+        853978956621483129u64,
+    ]));
+    /// COEFF_B = 0x9D8F71EEC83A44C3A1FBCEC6F5418E5C6154C2682B8AC231C5A3725C8170AAD
+    const COEFF_B: Fq = field_new!(Fq, BigInteger256([
+        7239382437352637935u64,
+        14509846070439283655u64,
+        5083066350480839936u64,
+        1265663645916442191u64,
+    ]));
+
+    type TEModelParameters = EdwardsParameters;
 }
 
 impl FromStr for EdwardsAffine {
--- a/algebra/src/curves/edwards_sw6/mod.rs
+++ b/algebra/src/curves/edwards_sw6/mod.rs
@@ -2,7 +2,7 @@ use crate::field_new;
 use crate::{
     biginteger::BigInteger384 as BigInteger,
     curves::{
-        models::{ModelParameters, TEModelParameters},
+        models::{ModelParameters, TEModelParameters, MontgomeryModelParameters},
         twisted_edwards_extended::{GroupAffine, GroupProjective},
     },
     fields::edwards_sw6::{fq::Fq, fr::Fr},
@@ -62,11 +62,36 @@ impl TEModelParameters for EdwardsParameters {
     /// AFFINE_GENERATOR_COEFFS = (GENERATOR_X, GENERATOR_Y)
     const AFFINE_GENERATOR_COEFFS: (Self::BaseField, Self::BaseField) = (GENERATOR_X, GENERATOR_Y);
 
+    type MontgomeryModelParameters = EdwardsParameters;
+
     /// Multiplication by `a` is just negation.
     #[inline(always)]
     fn mul_by_a(elem: &Self::BaseField) -> Self::BaseField {
         -*elem
     }
+}
+
+impl MontgomeryModelParameters for EdwardsParameters {
+    /// COEFF_A = 0x95D53EB3F6AC3F7A53C26020144439DC6073BCAE513E03FD06B6B3BAA390F25E51534B26719E33F4CD906D4DA9B535
+    const COEFF_A: Fq = field_new!(Fq, BigInteger([
+        7594254284108454966u64,
+        14287343397973578077u64,
+        6490358977072726023u64,
+        8023375322051995268u64,
+        8242802613686040715u64,
+        100541941146122331u64,
+    ]));
+    /// COEFF_B = 0x118650763CE64AB4BE743604C8D05013DC2663652A3D58B21ECAB7BFF65B70DB8BA09F9098E61CC903B2F92B2564ACA
+    const COEFF_B: Fq = field_new!(Fq, BigInteger([
+        11173793475516310780u64,
+        14217481814129454913u64,
+        11878518835804377107u64,
+        14866315431314324110u64,
+        9234787938768687129u64,
+        62053599622152261u64,
+    ]));
+
+    type TEModelParameters = EdwardsParameters;
 }
 
 impl FromStr for EdwardsAffine {
--- a/algebra/src/curves/jubjub/mod.rs
+++ b/algebra/src/curves/jubjub/mod.rs
@@ -2,7 +2,7 @@ use crate::field_new;
 use crate::{
     biginteger::BigInteger256,
     curves::{
-        models::{ModelParameters, TEModelParameters},
+        models::{ModelParameters, TEModelParameters, MontgomeryModelParameters},
         twisted_edwards_extended::{GroupAffine, GroupProjective},
     },
     fields::jubjub::{fq::Fq, fr::Fr},
@@ -89,11 +89,32 @@ impl TEModelParameters for JubJubParameters {
     /// AFFINE_GENERATOR_COEFFS = (GENERATOR_X, GENERATOR_Y)
     const AFFINE_GENERATOR_COEFFS: (Self::BaseField, Self::BaseField) = (GENERATOR_X, GENERATOR_Y);
 
+    type MontgomeryModelParameters = JubJubParameters;
+
     /// Multiplication by `a` is simply negation here.
     #[inline(always)]
     fn mul_by_a(elem: &Self::BaseField) -> Self::BaseField {
         -(*elem)
     }
+}
+
+impl MontgomeryModelParameters for JubJubParameters {
+    /// COEFF_A = 0xA002
+    const COEFF_A: Fq = field_new!(Fq, BigInteger256([
+        388496971701930u64,
+        6855257088226130262u64,
+        553476580979119549u64,
+        6516741293351590684u64,
+    ]));
+    /// COEFF_B = 0x73EDA753299D7D483339D80809A1D80553BDA402FFFE5BFEFFFFFFFEFFFF5FFD
+    const COEFF_B: Fq = field_new!(Fq, BigInteger256([
+        18446355550968045916u64,
+        10902955289292811939u64,
+        3147092737149958754u64,
+        6710871716016002197u64,
+    ]));
+
+    type TEModelParameters = JubJubParameters;
 }
 
 impl FromStr for JubJubAffine {
--- a/crypto-primitives/src/crh/bowe_hopwood/constraints.rs
+++ b/crypto-primitives/src/crh/bowe_hopwood/constraints.rs
@@ -1,0 +1,206 @@
+use algebra::Field;
+use std::hash::Hash;
+
+use crate::crh::{
+    bowe_hopwood::{BoweHopwoodPedersenCRH, BoweHopwoodPedersenParameters, CHUNK_SIZE},
+    pedersen::PedersenWindow,
+    FixedLengthCRHGadget,
+};
+use algebra::groups::Group;
+use r1cs_core::{ConstraintSystem, SynthesisError};
+use r1cs_std::{alloc::AllocGadget, groups::GroupGadget, uint8::UInt8};
+
+use r1cs_std::bits::boolean::Boolean;
+use std::{borrow::Borrow, marker::PhantomData};
+
+#[derive(Derivative)]
+#[derivative(Clone(
+    bound = "G: Group, W: PedersenWindow, ConstraintF: Field, GG: GroupGadget<G, ConstraintF>"
+))]
+pub struct BoweHopwoodPedersenCRHGadgetParameters<
+    G: Group,
+    W: PedersenWindow,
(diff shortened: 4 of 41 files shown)


crypto-primitives/src/crh/bowe_hopwood/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+impl<G: Group, W: PedersenWindow> FixedLengthCRH for BoweHopwoodPedersenCRH<G, W> {
+    const INPUT_SIZE_BITS: usize = PedersenCRH::<G, W>::INPUT_SIZE_BITS;
+    type Output = G;
+    type Parameters = BoweHopwoodPedersenParameters<G>;
+
+    fn setup<R: Rng>(rng: &mut R) -> Result<Self::Parameters, Error> {
+        fn calculate_num_chunks_in_segment<G: Group>() -> usize {
```
Pratyush:
```suggestion
        fn calculate_num_chunks_in_segment<F: PrimeField>() -> usize {
```

crypto-primitives/src/crh/bowe_hopwood/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+            let upper_limit = G::ScalarField::modulus_minus_one_div_two();
```
Pratyush:
```suggestion
            let upper_limit = F::modulus_minus_one_div_two();
```

crypto-primitives/src/crh/bowe_hopwood/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+            let mut c = 0;
+            let mut range = <G::ScalarField as PrimeField>::BigInt::from(2_u64);
```
Pratyush:
```suggestion
            let mut range = F::BigInt::from(2_u64);
```

crypto-primitives/src/crh/bowe_hopwood/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+            while range < upper_limit {
+                range.muln(4);
+                c += 1;
+            }
+
+            c
+        }
+
+        let maximum_num_chunks_in_segment = calculate_num_chunks_in_segment::<G>();
```
Pratyush:
```suggestion
        let maximum_num_chunks_in_segment = calculate_num_chunks_in_segment::<G::ScalarField>();
```

crypto-primitives/src/crh/bowe_hopwood/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+        ));
+        let generators = Self::create_generators(rng);
+        end_timer!(time);
+        Ok(Self::Parameters {
+            generators
+        })
+    }
+
+    fn evaluate(parameters: &Self::Parameters, input: &[u8]) -> Result<Self::Output, Error> {
+        let eval_time = start_timer!(|| "PedersenCRH::Eval");
```
Pratyush:
```suggestion
        let eval_time = start_timer!(|| "BoweHopwoodPedersenCRH::Eval");
```

r1cs-std/src/fields/mod.rs
```diff
@@ -38,6 +39,8 @@ pub trait FieldGadget<F: Field, ConstraintF: Field>:
 
     fn one<CS: ConstraintSystem<ConstraintF>>(_: CS) -> Result<Self, SynthesisError>;
 
+    fn add_bool_with_coeff<CS: ConstraintSystem<ConstraintF>>(&self, _: CS, _: &Boolean, _: F) -> Result<Self, SynthesisError>;
```
Pratyush: Maybe rename this to `conditionally_add_constant`? Currently it feels like you're adding the bool value to `self`.

r1cs-std/src/groups/curves/twisted_edwards/mod.rs
```diff
@@ -15,6 +15,206 @@ pub mod jubjub;
...
+        pub fn new(x: F, y: F) -> Self {
+            Self {
+                x,
+                y,
+                _params: PhantomData,
+                _engine: PhantomData,
+            }
+        }
+
+        pub fn params() -> MontgomeryParams<P> {
```
Pratyush: Longer term it could be nice to have a `MontgomeryParameters` trait (like the `TEModelParameters` trait), so that the `A` and `B` values generated here can be fixed constants.

r1cs-std/src/groups/curves/twisted_edwards/mod.rs
```diff
@@ -757,6 +957,122 @@ mod projective_impl {
...
+                    }
+                    Some(ref mut edwards_result) => {
+                        *edwards_result = GroupGadget::<TEAffine<P>, ConstraintF>::add(
+                            &segment_result,
+                            cs.ns(|| "edwards addition of leftover"),
+                            edwards_result,
+                        )?;
+                    }
+                }
+            }
```
Pratyush: These pieces of code are almost exactly the same, so it might be good to extract it into a macro or function/closure that is local to `precomputed_base_scalar_mul_3_bit_with_conditional_negation`?

r1cs-std/src/groups/mod.rs
```diff
@@ -111,6 +111,21 @@ pub trait GroupGadget<G: Group, ConstraintF: Field>:
         Ok(())
     }
 
+    fn precomputed_base_scalar_mul_3_bit_with_conditional_negation<'a, CS, T, I, B>(
```
Pratyush: From what I understand, the only difference between `precomputed_base_scalar_mul_3_bit_with_conditional_negation` and `precomputed_base_multiscalar_mul` is that the former requires an additional `segment_size` argument. Is there a way to avoid duplication in the interface itself?

Couple of options:
1) Have just one generic `precomputed_base_multiscalar_mul` whose signature is the same as `precomputed_base_scalar_mul_3_bit_with_conditional_negation`, but whose default implementation is the same as the current `precomputed_base_multiscalar_mul`. Then, for `TEGadget`, we just use the current implementation of `precomputed_base_scalar_mul_3_bit_with_conditional_negation` as the impl.
2) Investigate whether we actually need the `segment_size` argument, or whether it can be a fixed constant derivable from things like the scalar field size? I understand that, as used in the BHP hash, `segment_size` is just the window-size; is this value expected to differ across instantiations, or will it remain the same?

Pratyush (changes requested):
Thanks for the PR! Left a couple of quick typo fix suggestions, as well as a couple of questions regarding some code structure/naming.

r1cs-std/src/groups/mod.rs
kobigurk: That might be problematic - precomputed_base_scalar_mul_3_bit_with_conditional_negation doesn't have the same goal as precomputed_base_multiscalar_mul - it's more efficient, though it maps ([b1, b2, b3], G) to [-4,...,-1,1,...4]*G rather than [0,...,7]*G as precomputed_base_multiscalar_mul does.

I think we might be able to get rid of segment_size though.

r1cs-std/src/fields/mod.rs
kobigurk: Sounds great.

r1cs-std/src/groups/curves/twisted_edwards/mod.rs
kobigurk: Good idea, I added it.
It does require every Edwards curve to provide the Montgomery parameters.

r1cs-std/src/groups/mod.rs
kobigurk: Not 100% sure how to cleanly get rid of segment size yet. Let me know if you got ideas.

r1cs-std/src/groups/curves/twisted_edwards/mod.rs
kobigurk: Take a look at the latest commit.

r1cs-std/src/groups/mod.rs
kobigurk: The reason it's not direct is because it's not directly a property of the group when to move on to the next segment. The maximum is a property though. The question is whether to always use the maximum.

Pratyush: Re: merging the two, then maybe it might be worthwhile to rename the method to something different? To avoid confusion with normal multi-scalar multiplication. Do you have any ideas? (`precomputed_base_scalar_mul_3_bit_with_conditional_negation` is also a bit wordy)

If we rename it, then maybe the segment_size question is also moot, if we make it clearer what it means?. Another approach would be to potentially derive `segment_size` from the input length + group size? Do you have any references I could look at that talk about the derivation of segment_size?

kobigurk: That's a good idea - let's rename it. I'll think of something better.

The derivation of the maximum segment size is both implemented in https://github.com/celo-org/zexe/blob/052e66bb6153383f164aba8a3847f0e13bb2e903/crypto-primitives/src/crh/bowe_hopwood/mod.rs#L57 and described in the Zcash spec (https://github.com/zcash/zips/blob/master/protocol/protocol.pdf) in page 56.

Basically, it's the amount of generators you can use in this method and still not pass the maximum scalar.

If we always use the maximum one, then no doubt we can derive it. I think we might want to do that, there shouldn't be any real advantage to doing otherwise.

Pratyush:
The PR looks good, modulo the question regarding `precomputed_base_scalar_mul_3_bit_with_conditional_negation`, and the following point:

Instead of adding the `P: MontgomeryModelParameters` bound everywhere, do you think it would be better to add an associated type to `TEModelParameters` as follows:

```rust
pub trait TEModelParameters: ModelParameters {
	type MontModelParameters: MontgomeryParameters<BaseField == Self::BaseField> + From<Self>;
    // Rest here ...
}
```
and similarly in `MontgomeryModelParameters`:

```rust
pub trait MontgomeryModelParameters: ModelParameters {
	type TEModelParameters: TEModelParameters<BaseField == Self::BaseField> + From<Self>;
    // Rest here ...
}
```
This way one can just refer to `P::MontModelParameters` in the relevant places, while reducing the duplicated trait bounds. It would also allow us to rewrite `<P as TEModelParameters>::COEFF_A` to `P::COEFF_A`, which is slightly cleaner.

Let me know what you think; we can also land as-is and I can refactor as needed

kobigurk:
> The PR looks good, modulo the question regarding `precomputed_base_scalar_mul_3_bit_with_conditional_negation`, and the following point:
> 
> Instead of adding the `P: MontgomeryModelParameters` bound everywhere, do you think it would be better to add an associated type to `TEModelParameters` as follows:
> 
> ```rust
> pub trait TEModelParameters: ModelParameters {
> 	type MontModelParameters: MontgomeryParameters<BaseField == Self::BaseField> + From<Self>;
>     // Rest here ...
> }
> ```
> 
> and similarly in `MontgomeryModelParameters`:
> 
> ```rust
> pub trait MontgomeryModelParameters: ModelParameters {
> 	type TEModelParameters: TEModelParameters<BaseField == Self::BaseField> + From<Self>;
>     // Rest here ...
> }
> ```
> 
> This way one can just refer to `P::MontModelParameters` in the relevant places, while reducing the duplicated trait bounds. It would also allow us to rewrite `<P as TEModelParameters>::COEFF_A` to `P::COEFF_A`, which is slightly cleaner.
> 
> Let me know what you think; we can also land as-is and I can refactor as needed

That's a great idea! I think worth doing before merging.

r1cs-std/src/groups/mod.rs
Pratyush: I'll leave the decision regarding `segment_size` up to you! Re: name suggestions, maybe something based on `signed_digits`? Eg:

* `precomputed_base_signed_digit_scalar_mul`, with a comment explaining that it works only for 3-bit signed digits?
* `precomputed_base_3_bit_signed_digit_scalar_mul`

kobigurk:
Hey @Pratyush, 

I changed the structure of the code a bit - the responsibility on chunking the segments appropriately before getting to the `precomputed_base_3_bit_signed_digit_scalar_mul ` is now passed on to the caller - so there's no segment size needed anymore. The code looks cleaner overall, IMO.

Also made the change with the associated type.

algebra/src/curves/edwards_bls12/mod.rs
```diff
@@ -63,6 +65,25 @@ impl TEModelParameters for EdwardsParameters {
...
 
+impl MontgomeryModelParameters for EdwardsParameters {
+    /// COEFF_A = -1
+    const COEFF_A: Fq = field_new!(Fq, BigInteger256([
+        13800168384327121454u64,
+        6841573379969807446u64,
+        12529593083398462246u64,
+        853978956621483129u64,
+    ]));
+    /// COEFF_B = -1
```
Pratyush: Is this comment accurate?

algebra/src/curves/edwards_sw6/mod.rs
```diff
@@ -62,13 +62,38 @@ impl TEModelParameters for EdwardsParameters {
...
+    /// COEFF_A = -1
+    const COEFF_A: Fq = field_new!(Fq, BigInteger([
+        7594254284108454966u64,
+        14287343397973578077u64,
+        6490358977072726023u64,
+        8023375322051995268u64,
+        8242802613686040715u64,
+        100541941146122331u64,
+    ]));
+    /// COEFF_B = -1
```
Pratyush: Is the COEFF_B == -1?

algebra/src/curves/jubjub/mod.rs
```diff
@@ -89,13 +89,34 @@ impl TEModelParameters for JubJubParameters {
...
 
+impl MontgomeryModelParameters for JubJubParameters {
+    /// COEFF_A = -1
+    const COEFF_A: Fq = field_new!(Fq, BigInteger256([
+        388496971701930u64,
+        6855257088226130262u64,
+        553476580979119549u64,
+        6516741293351590684u64,
+    ]));
+    /// COEFF_B = -1
```
Pratyush: Same as the other COEFF_B comments

crypto-primitives/src/crh/bowe_hopwood/constraints.rs
```diff
@@ -0,0 +1,207 @@
...
+{
+    type OutputGadget = GG;
+    type ParametersGadget = BoweHopwoodPedersenCRHGadgetParameters<G, W, ConstraintF, GG>;
+
+    fn check_evaluation_gadget<CS: ConstraintSystem<ConstraintF>>(
+        cs: CS,
+        parameters: &Self::ParametersGadget,
+        input: &[UInt8],
+    ) -> Result<Self::OutputGadget, SynthesisError> {
+        // Pad the input if it is not the current length.
```
Pratyush, line 58:
```suggestion
        // Pad the input if it is not the correct length.
```

crypto-primitives/src/crh/bowe_hopwood/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+            W::WINDOW_SIZE,
+            W::NUM_WINDOWS,
+            CHUNK_SIZE,
+        );
+        for generators in parameters.generators.iter() {
+            assert_eq!(generators.len(), W::WINDOW_SIZE);
+        }
+        assert_eq!(CHUNK_SIZE, 3);
+
+        // Compute sum of h_i^{m_i} for all i.
```
Pratyush: Maybe update the comment to reflect the new mapping

Pratyush (changes requested):

Pratyush:
@kobigurk I left a couple of small comments, we can merge once those are addressed. Would you want to squash some of the commits? If not that's fine too.

algebra/src/curves/edwards_bls12/mod.rs
kobigurk: This was a placeholder 🤦‍♂

algebra/src/curves/edwards_sw6/mod.rs
kobigurk: Another placeholder.

kobigurk:
> @kobigurk I left a couple of small comments, we can merge once those are addressed. Would you want to squash some of the commits? If not that's fine too.

Alright, fixed those comments (oops), and squashed.

Pratyush:
Awesome, thanks so much for this contribution and for your patience with my reviews!

PR #37, opened 2019-10-20
Adds Groth16

kobigurk:


crypto-primitives/src/nizk/groth16/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+        end_timer!(proof_time);
+        Ok(result)
+    }
+
+    fn verify(
+        vk: &Self::PreparedVerificationParameters,
+        input: &Self::VerifierInput,
+        proof: &Self::Proof,
+    ) -> Result<bool, Error> {
+        let verify_time = start_timer!(|| "{Groth-Maller 2017}::Verify");
```
Pratyush:
```suggestion
        let verify_time = start_timer!(|| "{Groth 2016}::Verify");
```

crypto-primitives/src/nizk/groth16/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+        end_timer!(nizk_time);
+        Ok((pp, vk))
+    }
+
+    fn prove<R: Rng>(
+        pp: &Self::ProvingParameters,
+        input_and_witness: Self::AssignedCircuit,
+        rng: &mut R,
+    ) -> Result<Self::Proof, Error> {
+        let proof_time = start_timer!(|| "{Groth-Maller 2017}::Prove");
```
Pratyush:
```suggestion
        let proof_time = start_timer!(|| "{Groth 2016}::Prove");
```

crypto-primitives/src/nizk/groth16/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+        circuit: Self::Circuit,
+        rng: &mut R,
+    ) -> Result<
+        (
+            Self::ProvingParameters,
+            Self::PreparedVerificationParameters,
+        ),
+        Error,
+    > {
+        let nizk_time = start_timer!(|| "{Groth-Maller 2017}::Setup");
```
Pratyush:
```suggestion
        let nizk_time = start_timer!(|| "{Groth 2016}::Setup");
```

groth16/src/lib.rs
```diff
@@ -0,0 +1,300 @@
+//! An implementation of the [Groth][Groth16] simulation extractable zkSNARK.
```
Pratyush:
```suggestion
//! An implementation of the [Groth][Groth16] zkSNARK.
```

groth16/src/generator.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    end_timer!(h_time);
+
+    // Compute the L-query
+    let l_time = start_timer!(|| "Calculate L");
+    let mut l_query = FixedBaseMSM::multi_scalar_mul::<E::G1Projective>(
+        scalar_bits,
+        g1_window,
+        &g1_table,
+    &l,
```
Pratyush:
```suggestion
        &l,
```

Pratyush (changes requested):
Thanks for the PR, it looks great! Just a couple of additional comments:
- [x] Instead of adding the LICENSE files directly, could you instead make them symlinks like in the other crates?
- [x] Could you run the crate through `cargo fmt`? I noticed a couple of inconsistent indentation things.

kobigurk:
All done!

I noticed that `cargo fmt` with the latest nightly did more changes than I expected. Do you have a particular config you're using?

If not, we should run the whole Zexe workspace through it.

Pratyush:
I think it should use the rustfmt.toml that's in the repository root. From looking at the fmt commit, it seems that it used this config.

I usually run it on every crate except algebra because the formatting on constants gets screwed up by rustfmt.

kobigurk:
Got it. I think we can leave the formatting to another PR so this wouldn't affect the whole repository?

Pratyush:
Sure, as you wish!

Pratyush:
Thanks for the PR @kobigurk! It's great to have another implementation of Groth16 in Rust =)

Issue #50, opened 2020-01-07
`crates.io` release checklist


Pratyush:
`algebra`:

- [x] Move curve specific modules to their own crates. (Reduces compilation time)
- [x] Investigate how to make `ToBytes` and `FromBytes` `no_std` compatible (WIP in #76).
- [x] Possibly replace identity functions in `Field`, `Group`, and `{Projective,Affine}Curve` traits with `One`/`Zero` traits from `num_traits`. (Done in #54)
- [x] Possibly add `impl Add<Self> for Self` (besides just `impl Add<&Self> for Self>`). (Done in #53)
- [x] Add a generic trait for Point Compression for `AffineCurve` elements (See #51; done in #73).

`bench-utils`:
- [ ] Check whether it can be replaced/integrated with [`tracing`](https://github.com/tokio-rs/tracing).
- [ ] Move into `algebra` behind a `trace` feature?

`r1cs-core`:
- [x] Refactor `ConstraintSystem`, `ConstraintVar`, `LinearCombination`, and `Variable` APIs as per #34 and #43
- [ ] Rename `ConstraintSystem` to `R1CS`.
- [x] Prefer `ConstraintSystem::enter_ns` and `ConstraintSystem::leave_ns` to `ConstraintSystem::ns`. Additionally, refactor `Namespace` struct to be just a guard struct which automatically decrements the scope when dropped.

`r1cs-std`:
- [x] Make `FieldVar`, `GroupVar`, and other variable structs generic over `CS: ConstraintSystem<F>` as well. This enables getting rid of `cs: CS` arguments to various functions like `add`, `mul`, etc. In turn, this allows `impl Add<Self> for FieldVar<...>` impls.
- [x] impl new `impl {Add, Mul}<Self::Constant> for FieldVar<...>`, impls with the constants. (This might not be possible due to coherence rules, might have to use `enum`-based approach)
- [x] standardize `UInt{8, 16, 32, 64}`.
- [ ] Make derive macros so that traits like `EqGadget`, `ToBytesGadget` are less of a pain to implement.
- [x] Add support for non-native `FpGadget`s.

`groth16` + `gm17`:
- [x] Adopt similar code organization as [`marlin`](github.com/scipr-lab/marlin)
- [x] Implement proving key, verification key, and proof serialization (blocked on #51).

`crypto-primitives`:
- [x] Make gadget APIs consistent with above conventions.

kobigurk:
[...]

PR #53, opened 2020-01-13
Add `impl Add<Self> for Self`, `impl Mul<Self> for Self` for field mo…

huitseeker:
…dels

Contributes to #50.

/cc @Pratyush

### DIFF
--- a/algebra/src/curves/sw6/mod.rs
+++ b/algebra/src/curves/sw6/mod.rs
@@ -1,7 +1,7 @@
-use crate::field_new;
 use crate::{
     biginteger::BigInteger832,
     curves::{PairingCurve, PairingEngine},
+    field_new,
     fields::{
         sw6::{
             fq::{Fq, FqParameters},
@@ -146,7 +146,7 @@ impl SW6 {
         let mut elt_q3 = elt.clone();
         elt_q3.frobenius_map(3);
         // elt_q3_over_elt = elt^(q^3-1)
-        let elt_q3_over_elt = elt_q3 * &elt_inv;
+        let elt_q3_over_elt = elt_q3 * elt_inv;
         // alpha = elt^((q^3-1) * q)
         let mut alpha = elt_q3_over_elt.clone();
         alpha.frobenius_map(1);
--- a/algebra/src/fields/macros.rs
+++ b/algebra/src/fields/macros.rs
@@ -105,3 +105,25 @@ macro_rules! sqrt_impl {
         }
     }};
 }
+
+// Implements AddAssign on Self by deferring to an implementation on &Self
+macro_rules! impl_addassign_from_ref {
+    ($field: ident, $params: ident) => {
+        impl<P: $params> AddAssign<Self> for $field<P> {
+            fn add_assign(&mut self, other: Self) {
+                self.add_assign(&other)
+            }
+        }
+    };
+}
+
+// Implements MulAssign on Self by deferring to an implementation on &Self
+macro_rules! impl_mulassign_from_ref {
+    ($field: ident, $params: ident) => {
+        impl<P: $params> MulAssign<Self> for $field<P> {
+            fn mul_assign(&mut self, other: Self) {
+                self.mul_assign(&other)
+            }
+        }
+    };
+}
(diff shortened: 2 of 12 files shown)


algebra/src/curves/sw6/mod.rs
```diff
@@ -146,7 +146,7 @@ impl SW6 {
         let mut elt_q3 = elt.clone();
         elt_q3.frobenius_map(3);
         // elt_q3_over_elt = elt^(q^3-1)
-        let elt_q3_over_elt = elt_q3 * &elt_inv;
+        let elt_q3_over_elt = elt_q3 * elt_inv;
```
Pratyush, line 149: Was this failing to compile after you added the new impls?

Pratyush (commented):
Thanks for the PR! It looks good modulo the nit

algebra/src/curves/sw6/mod.rs
huitseeker: Yes, with:
```rust
error[E0277]: cannot multiply `&&fields::models::fp6_2over3::Fp6<fields::sw6::fq6::Fq6Parameters>` to `fields::models::fp6_2over3::Fp6<fields::sw6::fq6::Fq6Parameters>`
   --> algebra/src/curves/sw6/mod.rs:149:38
    |
149 |         let elt_q3_over_elt = elt_q3 * &elt_inv;
    |                                      ^ no implementation for `fields::models::fp6_2over3::Fp6<fields::sw6::fq6::Fq6Parameters> * &&fields::models::fp6_2over3::Fp6<fields::sw6::fq6::Fq6Parameters>`
    |
    = help: the trait `std::ops::Mul<&&fields::models::fp6_2over3::Fp6<fields::sw6::fq6::Fq6Parameters>>` is not implemented for `fields::models::fp6_2over3::Fp6<fields::sw6::fq6::Fq6Parameters>`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
```

Note that `elt_inv: &Fq6`, so that I think the trait bound is `&Fq6: Mul<&'a $0>`, which would not match in the presence of `impl<P: Fp6Parameters> Mul<Fp6<P>> for Fp6<P>`. IIUC, in its absence (i.e. before this commit), autoref would fire, leading the trait system to solve for `Fq6: Mul<&'a $0>` which matches the (single) existing `impl Mul`.

(Yes, I'm praying for the success of rust-lang issue 44619)

Pratyush:
I probably should have asked this earlier, but it would be great to also have the corresponding `AddAssign` and `MulAssign` impls, maybe via blanket impls (`impl<F: Field> AddAssign<F> for F` where we just invoke the existing AddAssign impl). We can also add these in a followup PR if that's better.

huitseeker:
@Pratyush In this case, are you opposed to just using macros, à la [auto_impl](https://github.com/auto-impl-rs/auto_impl)? (I'll need to write something custom, as the derive trigger of `auto_impl` won't be available)

Pratyush:
Hmm wouldn’t it be possible to just write a blanket impl that just invokes the existing reference-based impls? Then you only need one impl block

huitseeker:
@Pratyush one impl block per instance, right? The trait `AddAssign` is non-local to `Field`, so that I'll get https://gist.github.com/huitseeker/bf2a822fdb06fe20bba29b3a82f0ea9f#file-gistfile1-txt-L2-L9 if I add a blanket impl to `field.rs`

Pratyush:
Ahh the orphan rule strikes again. Good point. Let's just do a macro, maybe something like what already exists in `{bigint,fields}/macros.rs`.

huitseeker:
@Pratyush Note I will need to remove a lot of references in `algebra/src/fields/models/` in the operands to the calls to `mul_assign`. If unmodified, those `self.c0.mul_assign(&element)` will indeed generate 
```rust
  --> algebra/src/fields/models/fp6_3over2.rs:70:28
   |
70 |         self.c0.mul_assign(&element);
   |                            ^^^^^^^^ no implementation for `fields::models::fp2::Fp2<<P as fields::models::fp6_3over2::Fp6Parameters>::Fp2Params> *= &&fields::models::fp2::Fp2<<P as fields::models::fp6_3over2::Fp6Parameters>::Fp2Params>`
```

.. for precisely the reason we explored [above](https://github.com/scipr-lab/zexe/pull/53#discussion_r366333475). There's [a ton](https://gist.github.com/b379430dfd2e6ffc8aa579ea41711b4e) of such instances, so I want to make sure we're on the same page first.

Pratyush:
Oops, just saw your last comment. Yeah, that’s fine, if anything it’s more correct now!

huitseeker:
@Pratyush Done!

Pratyush:
Awesome, thanks!

PR #54, opened 2020-01-14
Convert identity functions in Field, Group, and {Projective,Affine}Curve traits with One/Zero traits from num_traits.

huitseeker:
- contributes to #50,
- depends on #53 and builds on it,
- due ~~to coherence &~~ requirements of `num_traits::{Zero, One}` to implement
  `std::ops::Add<Self, ..>` and (resp.) `std::ops::Mul<Self, ..>`, I've had
  to ~~replace~~ [edit:complement] the afferent `impl<'a, P: ..> (Add|Mul)<&'a Self> for
  Group(Affine|Projective)<P>` ~~by~~ [edit:with] direct implementations on `Self`,
- I did not have to fight the borrow checker for this conversion => I think
  this hints arithmetic operations are called in contexts where the operand
  is owned,
- hence should this end up on a merge track, we may want to open an issue
  to convert the `impl<'a, P:..> (Neg|Sub|..)<&'a Self> for ..<P>` trait
  usage to direct `impl<P:..> (Neg|Sub|..)<Self> for ..<P>`
- ~~the `impl AddAssign for GroupAffine<P>` in
  curves/models/short_weierstrass_jacobian.rs is provided to fit trait
  bounds, and without any guarantee of suitability for any particular purpose~~
- ~~even though I don't think it's used.~~

### DIFF
--- a/algebra/src/curves/models/short_weierstrass_projective.rs
+++ b/algebra/src/curves/models/short_weierstrass_projective.rs
@@ -130,8 +130,8 @@ impl<P: Parameters> Add<Self> for GroupAffine<P> {
 impl<'a, P: Parameters> AddAssign<&'a Self> for GroupAffine<P> {
     fn add_assign(&mut self, other: &'a Self) {
         let lambda = (other.y - &self.y) / &(other.x - &self.y);
-        let x3 = lambda * lambda - &self.x - &other.x;
-        let y3 = (self.x - &x3) * lambda - &self.y;
+        let x3 = lambda * &lambda - &self.x - &other.x;
+        let y3 = (self.x - &x3) * &lambda - &self.y;
 
         self.x = x3;
         self.y = y3;
--- a/algebra/src/curves/models/short_weierstrass_projective.rs
+++ b/algebra/src/curves/models/short_weierstrass_projective.rs
@@ -241,7 +241,9 @@ impl<P: Parameters> PartialEq for GroupProjective<P> {
         // x1/z1 == x2/z2  <==> x1 * z2 == x2 * z1
         if (self.x * &other.z) != (other.x * &self.z) {
             false
-        } else { (self.y * &other.z) == (other.y * &self.z) }
+        } else {
+            (self.y * &other.z) == (other.y * &self.z)
+        }
     }
 }
 
@@ -493,6 +495,15 @@ impl<P: Parameters> Add<Self> for GroupProjective<P> {
     fn add(self, other: Self) -> Self {
         let mut copy = self;
         copy += &other;
+        copy
+    }
+}
+
+impl<'a, P: Parameters> Add<&'a Self> for GroupProjective<P> {
+    type Output = Self;
+    fn add(self, other: &'a Self) -> Self {
+        let mut copy = self;
+        copy += other;
         copy
     }
 }
--- a/algebra/src/curves/models/short_weierstrass_projective.rs
+++ b/algebra/src/curves/models/short_weierstrass_projective.rs
@@ -129,12 +129,9 @@ impl<P: Parameters> Add<Self> for GroupAffine<P> {
 
 impl<'a, P: Parameters> AddAssign<&'a Self> for GroupAffine<P> {
     fn add_assign(&mut self, other: &'a Self) {
-        let lambda = (other.y - &self.y) / &(other.x - &self.y);
-        let x3 = lambda * &lambda - &self.x - &other.x;
-        let y3 = (self.x - &x3) * &lambda - &self.y;
-
-        self.x = x3;
-        self.y = y3;
+        let mut s_proj = self.into_projective();
+        s_proj.add_assign_mixed(&other);
+        *self = s_proj.into_affine();
     }
 }
 
--- a/algebra/src/curves/models/short_weierstrass_projective.rs
+++ b/algebra/src/curves/models/short_weierstrass_projective.rs
@@ -1,6 +1,7 @@
 use crate::curves::models::SWModelParameters as Parameters;
 use rand::{Rng, distributions::{Standard, Distribution}};
 use crate::UniformRand;
+use num_traits::{One, Zero};
 use std::{
     fmt::{Display, Formatter, Result as FmtResult},
     io::{Read, Result as IoResult, Write},
@@ -107,14 +108,40 @@ impl<P: Parameters> GroupAffine<P> {
     }
 }
 
+impl<P: Parameters> Zero for GroupAffine<P> {
+    fn zero() -> Self {
+        Self::new(P::BaseField::zero(), P::BaseField::one(), true)
+    }
+
+    fn is_zero(&self) -> bool {
+        self.infinity
+    }
+}
+
+impl<P: Parameters> Add<Self> for GroupAffine<P> {
+    type Output = Self;
+    fn add(self, other: Self) -> Self {
+        let mut copy = self;
+        copy += &other;
+        copy
+    }
+}
+
+impl<'a, P: Parameters> AddAssign<&'a Self> for GroupAffine<P> {
+    fn add_assign(&mut self, other: &'a Self) {
+        let lambda = (other.y - &self.y) / &(other.x - &self.y);
+        let x3 = lambda * lambda - &self.x - &other.x;
+        let y3 = (self.x - &x3) * lambda - &self.y;
+
+        self.x = x3;
+        self.y = y3;
+    }
+}
+
 impl<P: Parameters> AffineCurve for GroupAffine<P> {
     type BaseField = P::BaseField;
     type ScalarField = P::ScalarField;
     type Projective = GroupProjective<P>;
-
-    fn zero() -> Self {
-        Self::new(Self::BaseField::zero(), Self::BaseField::one(), true)
-    }
 
     fn prime_subgroup_generator() -> Self {
         Self::new(
@@ -122,10 +149,6 @@ impl<P: Parameters> AffineCurve for GroupAffine<P> {
             P::AFFINE_GENERATOR_COEFFS.1,
             false,
         )
-    }
-
-    fn is_zero(&self) -> bool {
-        self.infinity
     }
 
     fn mul<S: Into<<Self::ScalarField as PrimeField>::BigInt>>(&self, by: S) -> GroupProjective<P> {
@@ -231,9 +254,6 @@ impl<P: Parameters> Distribution<GroupProjective<P>> for Standard {
     }
 }
 
-
-
-
 impl<P: Parameters> ToBytes for GroupProjective<P> {
     #[inline]
     fn write<W: Write>(&self, mut writer: W) -> IoResult<()> {
@@ -271,11 +291,7 @@ impl<P: Parameters> GroupProjective<P> {
     }
 }
 
-impl<P: Parameters> ProjectiveCurve for GroupProjective<P> {
-    type BaseField = P::BaseField;
-    type ScalarField = P::ScalarField;
-    type Affine = GroupAffine<P>;
-
+impl<P: Parameters> Zero for GroupProjective<P> {
     // The point at infinity is always represented by Z = 0.
     #[inline]
     fn zero() -> Self {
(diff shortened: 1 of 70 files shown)


Pratyush:
Thanks for the PR. The removal of the `impl<'a, P...> Add<&'a Self> for Self` is a bit unfortunate. Is there no way to have both impls?

Also, would it be possible to split out commit into 2 commits:
* The first commit would have just the new `Zero/One` commit.
* The second commit would have the changes resulting from removing the `&` everywhere,

huitseeker:
- [`One`](https://docs.rs/num-traits/0.2.11/num_traits/identities/trait.One.html) requires the bound `Mul<Self, Output = Self>`, so adding `One` and adding the `impl<P:..> Mul<Self> for Foo<P>` have to be in the same commit,
- [`Zero`](https://docs.rs/num-traits/0.2.11/num_traits/identities/trait.Zero.html) requires the bound `Add<Self, Output = Self>`, so adding `Zero` and adding the `impl<P:..> Add<Self> for Foo<P>` have to be in the same commit,
- removing the `&` everywhere has to be in the same commit as adding the `impl<P:..> (Add|Mul)<Self> for Foo<P>` because the presence of `(Add|Mul)` for a callee, rather than for its reference, changes trait dispatch (autoref/deref has lower priority). To wit, I've uploaded [a branch](https://github.com/huitseeker/zexe/tree/num_traits_2) with the `num_trait` additions but without changing (de-)references, where we can witness this sort of error:
```rust
error[E0308]: mismatched types                                                                                                                                                                                                                                                   
  --> algebra/src/curves/models/short_weierstrass_jacobian.rs:81:44                                                                                                                                                                                                              
   |                                                                                                                                                                                                                                                                             
81 |         let x3b = P::add_b(&((x.square() * &x) + &P::mul_by_a(&x)));                                                                                                                                                                                                        
   |                                            ^^                                                                                                                                                                                                                               
   |                                            |                                                                                                                                                                                                                                
   |                                            expected associated type, found reference                                                                                                                                                                                        
   |                                            help: consider removing the borrow: `x`                                                                                                                                                                                          
   |                                                                                                                                                                                                                                                                             
   = note: expected associated type `<P as curves::models::ModelParameters>::BaseField`                                                                                                                                                                                          
                    found reference `&<P as curves::models::ModelParameters>::BaseField`            
```

 
- Thanks for the pushback on the removed `impl<'a, P:..> Add<&'a Self> for GroupProjective<P>`, there was indeed no coherence problem, I re-added those to the last commit of this PR.
- did you have a chance to look at `impl<P:..> Add<Self> for GroupAffine<P>` in `short_weierstrass_jacobian`? The trait bounds on `Zero` force me to implement it.

algebra/src/curves/models/short_weierstrass_projective.rs
```diff
@@ -107,15 +110,41 @@ impl<P: Parameters> GroupAffine<P> {
...
+impl<P: Parameters> Add<Self> for GroupAffine<P> {
+    type Output = Self;
+    fn add(self, other: Self) -> Self {
+        let mut copy = self;
+        copy += &other;
+        copy
+    }
+}
+
+impl<'a, P: Parameters> AddAssign<&'a Self> for GroupAffine<P> {
```
Pratyush, line 130: Maybe this impl can be replaced with something like `self.into_projective().add_assign_mixed(&other)` (or whichever variant compiles)

This would reduce some duplication.

algebra/src/fields/mod.rs
```diff
@@ -64,37 +65,27 @@ pub trait Field:
...
     + UniformRand
+    + Zero
     + Sized
     + Hash
     + From<u128>
     + From<u64>
     + From<u32>
     + From<u16>
     + From<u8>
-    + for<'a> Add<&'a Self, Output = Self>
```
Pratyush, line 77: Do we need to remove these bounds?

algebra/src/fields/mod.rs
```diff
@@ -64,37 +65,27 @@ pub trait Field:
...
-    + for<'a> Mul<&'a Self, Output = Self>
+    + Mul<Self, Output = Self>
     + for<'a> Div<&'a Self, Output = Self>
     + for<'a> AddAssign<&'a Self>
     + for<'a> SubAssign<&'a Self>
     + for<'a> MulAssign<&'a Self>
     + for<'a> DivAssign<&'a Self>
 {
-    /// Returns the zero element of the field, the additive identity.
-    fn zero() -> Self;
```
Pratyush, line 87: Maybe it might make sense to keep these functions and have a default implementation that just invokes the methods from `Zero` and `One`? This would avoid having to type  `use num_traits::{Zero, One};` everywhere. (Ditto for the curve and group traits)

algebra/src/fields/mod.rs
Pratyush: In particular, the following error in https://github.com/scipr-lab/zexe/pull/54#issuecomment-574195828 could be because these bounds were removed, and so `rustc` errored out.
```rust
error[E0308]: mismatched types                                                                                                                                                                                                                                                   
  --> algebra/src/curves/models/short_weierstrass_jacobian.rs:81:44                                                                                                                                                                                                              
   |                                                                                                                                                                                                                                                                             
81 |         let x3b = P::add_b(&((x.square() * &x) + &P::mul_by_a(&x)));                                                                                                                                                                                                        
   |                                            ^^                                                                                                                                                                                                                               
   |                                            |                                                                                                                                                                                                                                
   |                                            expected associated type, found reference                                                                                                                                                                                        
   |                                            help: consider removing the borrow: `x`                                                                                                                                                                                          
   |                                                                                                                                                                                                                                                                             
   = note: expected associated type `<P as curves::models::ModelParameters>::BaseField`                                                                                                                                                                                          
                    found reference `&<P as curves::models::ModelParameters>::BaseField`  
```

Pratyush:
(The primary reason I'm worried about removing the `&`s everywhere is because it might have a performance impact)

algebra/src/fields/mod.rs
huitseeker: Please consider https://gist.github.com/1e4f5aaf15f910a53896f4dfb9f68627 which is a compilation log on my branch [num_traits_2](https://github.com/huitseeker/zexe/tree/num_traits_2) at [fe83e42](https://github.com/huitseeker/zexe/commit/fe83e42). As we discussed, this branch does not contain any of the reference adjustments, but does contain the traits bounds above, along with everything else on this PR.

(But fair point on the removed trait bounds, I've reintroduced them to the present PR).

huitseeker:
@Pratyush The one thing we can do modularity-wise is to put the reference removal and the addition of the direct `impl (Add|Mul)`, and leave the introduction of `Zero` and `Mul` to a later commit.

algebra/src/fields/mod.rs
huitseeker: Note that in many places, when we add the `use` for the trait, we remove the one to `Field`.

The risk I see in your suggestion is that the inherent method would be chosen ([in most cases](https://doc.rust-lang.org/reference/expressions/method-call-expr.html)) over the trait method, whereas the inherent method would not convey much more information. 
- is the `Zero` / `Mul` implementation still relevant then?
- is there a chance some future developer makes a modification to the trait implementation and forgets to apply it to the inherent method?

A way to not have to edit those `use` statements is maybe ... just to not use the `num_traits`.

huitseeker:
@Pratyush I've submitted #58 in the general direction of assuaging those performance impact worries. The present PR does not contain any added `.clone()` but I do better see where the prevalence of `Copy` would leave you worried, now.

algebra/src/fields/mod.rs
Pratyush: Hmm I was able to get it to compile with the following patch:

```
From e1c777280d6956c4dbf9015d34072e85c762ac80 Mon Sep 17 00:00:00 2001
From: Pratyush Mishra <pratyushmishra@berkeley.edu>
Date: Wed, 15 Jan 2020 08:06:21 -0800
Subject: [PATCH] Compile with refs

---
 algebra/src/curves/models/short_weierstrass_jacobian.rs   | 4 ++--
 algebra/src/curves/models/short_weierstrass_projective.rs | 4 ++--
 algebra/src/fields/mod.rs                                 | 4 +---
 3 files changed, 5 insertions(+), 7 deletions(-)

diff --git a/algebra/src/curves/models/short_weierstrass_jacobian.rs b/algebra/src/curves/models/short_weierstrass_jacobian.rs
index be4924a..318c33e 100644
--- a/algebra/src/curves/models/short_weierstrass_jacobian.rs
+++ b/algebra/src/curves/models/short_weierstrass_jacobian.rs
@@ -129,8 +129,8 @@ impl<P: Parameters> Add<Self> for GroupAffine<P> {
 impl<'a, P: Parameters> AddAssign<&'a Self> for GroupAffine<P> {
     fn add_assign(&mut self, other: &'a Self) {
         let lambda = (other.y - &self.y) / &(other.x - &self.y);
-        let x3 = lambda * lambda - &self.x - &other.x;
-        let y3 = (self.x - &x3) * lambda - &self.y;
+        let x3 = lambda * &lambda - &self.x - &other.x;
+        let y3 = (self.x - &x3) * &lambda - &self.y;
 
         self.x = x3;
         self.y = y3;
diff --git a/algebra/src/curves/models/short_weierstrass_projective.rs b/algebra/src/curves/models/short_weierstrass_projective.rs
index 1602a35..1ab87cc 100644
--- a/algebra/src/curves/models/short_weierstrass_projective.rs
+++ b/algebra/src/curves/models/short_weierstrass_projective.rs
@@ -132,8 +132,8 @@ impl<P: Parameters> Add<Self> for GroupAffine<P> {
 impl<'a, P: Parameters> AddAssign<&'a Self> for GroupAffine<P> {
     fn add_assign(&mut self, other: &'a Self) {
         let lambda = (other.y - &self.y) / &(other.x - &self.y);
-        let x3 = lambda * lambda - &self.x - &other.x;
-        let y3 = (self.x - &x3) * lambda - &self.y;
+        let x3 = lambda * &lambda - &self.x - &other.x;
+        let y3 = (self.x - &x3) * &lambda - &self.y;
 
         self.x = x3;
         self.y = y3;
diff --git a/algebra/src/fields/mod.rs b/algebra/src/fields/mod.rs
index 37d0391..a49c322 100644
--- a/algebra/src/fields/mod.rs
+++ b/algebra/src/fields/mod.rs
@@ -77,11 +77,9 @@ pub trait Field:
     + From<u32>
     + From<u16>
     + From<u8>
-    + for<'a> Add<&'a Self, Output = Self>
     + Add<Self, Output = Self>
     + for<'a> Add<&'a Self, Output = Self>
     + for<'a> Sub<&'a Self, Output = Self>
-    + for<'a> Mul<&'a Self, Output = Self>
     + Mul<Self, Output = Self>
     + for<'a> Mul<&'a Self, Output = Self>
     + for<'a> Div<&'a Self, Output = Self>
@@ -341,7 +339,7 @@ pub fn batch_inversion<F: Field>(v: &mut [F]) {
         .zip(prod.into_iter().rev().skip(1).chain(Some(F::one())))
     {
         // tmp := tmp * g.z; g.z := tmp * s = 1/z
-        let newtmp = tmp * &f;
+        let newtmp = tmp * *f;
         *f = tmp * &s;
         tmp = newtmp;
     }
-- 
2.17.1
```

algebra/src/fields/mod.rs
Pratyush: The primary reason I want the  `num_traits` impls stuff is for compatibility with generic code (such as the `ndarray` crate) that relies on these traits. 

To resolve both our concerns, maybe it suffices to keep the one and zero methods on the Field/Group traits, and then do a blanket impl:

```
impl<F: Field> One for F {
	fn one() -> F {
		<F as Field>::one()
	}
}
```

This prevents downstream developers from overriding the correct implementation, while still having the generic impls.

algebra/src/fields/mod.rs
huitseeker: This takes a tad more work than your patch (which does the work for the algebra module), but happy to take things in this direction. Stand by.

algebra/src/fields/mod.rs
huitseeker: I'd love this to be the case, but the traits `One` / `Zero` are non-local to zexe, so the blanket implementation on `Field`-bounded types is running afoul of orphan rules, I think.

algebra/src/fields/mod.rs
Pratyush: Awesome, thanks so much for your patience =)

algebra/src/fields/mod.rs
Pratyush: Bleh again the orphan rules lol. Ok, your current approach seems the best way to proceed; maybe we can re-export the traits from the algebra prelude to make life easier.

algebra/src/curves/models/short_weierstrass_projective.rs
huitseeker: Great suggestion, addressed in 02a3685

Pratyush: Awesome, thanks!

huitseeker:
@Pratyush Once we manage to get this one merged (let me know if you have more feedback!) I have lined up a [derive macro](https://github.com/huitseeker/zexe/compare/num_traits...huitseeker:derive_macros?expand=1) improvement, that should remove a lot of the manual management of declaring arithmetic`Foo<Self>` implementations that defer to `Foo<&Self>`.

Pratyush:
Awesome, is it ready for a final review?

huitseeker:
@Pratyush #54 is, yeah!

Pratyush:
Ok, thanks for the PR and the patience!

PR #58, opened 2020-01-15
Remove unused / uneeded `Copy` instances in algebra

huitseeker:
This should help us build more confidence in avoiding accidental
performance-impacting copies.

### DIFF
--- a/algebra/src/curves/bls12_377/g1.rs
+++ b/algebra/src/curves/bls12_377/g1.rs
@@ -8,7 +8,7 @@ use crate::{
     },
 };
 
-#[derive(Copy, Clone, Default, PartialEq, Eq)]
+#[derive(Clone, Default, PartialEq, Eq)]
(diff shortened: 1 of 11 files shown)


Pratyush:
Sorry this slipped under my radar. This is a nice clean up, but I'd like to confirm that we didn't accidentally make any of `AffineCurve` or `ProjectiveCurve` impls non-`Copy`; is it possible to add tests that fail to compile if the relevant structs are non-`Copy`?

huitseeker:
@Pratyush So, I can put in such a static check as soon as the repo contains a struct that implements `(Affine|Projective)Curve`. The struct has to be fully instantiated, no generic parameters left.

Pratyush:
Actually, I just realized that we don't need to worry about this, because the `Field`/`Group`/`*Curve` traits all require implementers to be `Copy`.

PR #59, opened 2020-01-16
r1cs-std: makes extension field to_bits create namespaces for the ind…

kobigurk:
…ividual elements

Otherwise, when using `to_bits`, there are conflicts with the variable names.


Pratyush:
Oops, thanks for the PR!

PR #60, opened 2020-01-16
Fix upcoming deprecations

Pratyush:


huitseeker:
This LGTM. I would also grealty appreciate a commit of a run of rustfmt once we're done with the current batch of PRs :)

Pratyush:
Yes, a rustfmt run would be good, but before that I need to figure out how to make rustfmt not screw up formatting of all the field constants that we use.

huitseeker:
@Pratyush PFA I've submitted https://github.com/scipr-lab/zexe/pull/61 towards that objective. After that `rustfmt` does not touch numerical constants any more.

PR #61, opened 2020-01-16
Exempt most numerical constants from formatting

huitseeker:
Uses `#[rustfmt::skip]` annotations on field constants and numerical definitions.


huitseeker:
@Pratyush After this, we should be able to turn on rustfmt on the code base.

Pratyush:
Hmm is there no way to selectively a given rule (which reformats arrays) for an entire file? Otherwise this is a lot of annotation (and somewhat easy to miss when adding new curves)

(I'm sorry that you had to add so many annotations!)

huitseeker:
@Pratyush AFAIK, this is [not stable](https://github.com/rust-lang/rustfmt/issues/3243). besides, there are completely legitimate formatting errors in most files that `rustfmt` should run on to fix. Finally, most of the changes in this PR are tool-powered, I even quote the one-liners explicitly in some commits.

huitseeker:
@Pratyush Or maybe there is..  what do you think about the Visual vs Block difference here?
https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#array

algebra/src/fields/bls12_377/fq2.rs
```diff
@@ -16,6 +16,8 @@ impl Fp2Parameters for Fq2Parameters {
     type Fp = Fq;
 
     /// NONRESIDUE = -5
+    #[rustfmt::skip]
+    #[rustfmt::skip]
```
Pratyush:
```suggestion
```

Pratyush (commented):
This approach is fine as is; I don't think we can do significantly better at this point in time. Let's merge once the small nit is applied and after rebasing =)

huitseeker:
@Pratyush Done!

Pratyush:
Awesome, thanks!

PR #67, opened 2020-01-19
Correct some documentation in algebra

ValarDragon:
(Thanks Pratyush for the help w/ this)

FpParameter's Inv:

The current comment is incorrect.
The updated comment should be correct independent of hardware. This is because the integers used are explicitly declared as u64 throughout the library.

I've confirmed this equation matches what is in edwards_bls12.

Edwards_bls12:
Fix the mismatch between comment and generator

Field generator:
Remove redundant information about it being a quadratic non-residue. 

A quadratic residue is the square of an element. As we are in a non-binary field, the order of the multiplicative subgroup is even. Therefore any element with a quadratic residue cannot be a multiplicative generator.


Pratyush:
thanks for the PR!

weikengchen:
It seems that the new comment about edwards_bls12 is still incorrect (the u64's are in reverse order).

It is actually 514816350737603056810416736903601868499289492298974714218162527411995454738

I may push a PR later to fix this.

weikengchen:
Aha, it is 5 in the Montgomery reduction form.

jon-chuang:
Can I clarify that edwards_bls12 is an edwards curve used for ECDSA etc. inside BLS12 (inside a SNARK)? Actually, which BLS12 curve's subgroup order is its prime modulus equal to then, 377/381? 

Is jubjub the corresponding Edwards curve for bls12_381? Why does the SW6 curve require an Edwards curve too?

weikengchen:
The edwards_* and jubjub are twisted Edwards curves targeting for Pedersen hash functions. Over the Edwards curve, some computations are more efficient and can lead to efficient hash function constructions in SNARK.

edwards_bls12 is under bls12_377.
jubjub is under bls12_381.

The Edwards curve under SW6 is used for Pedersen hash function, too. It is used to specifically create a CRH for the SNARK verifying key used by the verifier in the constraint system over SW6. The related code is here: https://github.com/scipr-lab/zexe/blob/master/dpc/src/dpc/plain_dpc/instantiated.rs

PR #73, opened 2020-01-24
Adds serialization

### LINKED ISSUE #51: Add point compression + canonical serialization for `AffineCurve` elements
This is one of the critical missing pieces for this library. A sketch of the design I have in mind is to make `CanonicalSerialize` and `CanonicalDeserialize` traits:
```rust
pub trait CanonicalSerialize {
	type Output: Iterator<Item = u8>
	fn serialize(&self) -> Output;
}

pub trait CanonicalDeserialize: CanonicalSerialize {
	fn deserialize(&Self::Output) -> Self;
}
```
The idea would then be to implement these directly for the final instantiated curves, e.g. for `Bls12_381_G1`. (I don't see an easy way to have a generic implementation that works for all base fields.)

kobigurk:
Closes #51.

### DIFF
--- a/algebra/src/serialize/error.rs
+++ b/algebra/src/serialize/error.rs
@@ -11,6 +11,8 @@ pub enum SerializationError {
     NotEnoughSpace,
     /// During serialization, the data was invalid.
     InvalidData,
+    /// During serialization, extra info was of the wrong size.
+    ExtraInfoWrongSize,
     /// During serialization, we countered an I/O error.
     IoError(io::Error),
 }
@@ -42,6 +44,9 @@ impl fmt::Display for SerializationError {
                 },
                 SerializationError::InvalidData => {
                     "the input buffer contained invalid data"
+                },
+                SerializationError::ExtraInfoWrongSize => {
+                    "extra info is of the wrong size"
                 },
                 SerializationError::IoError(_) => "encountered an I/O error",
             };
--- a/algebra/src/serialize/error.rs
+++ b/algebra/src/serialize/error.rs
@@ -1,0 +1,51 @@
+use std::fmt;
+use std::io;
+use std::error::Error;
+
+/// This is an error that could occur during serialization
+#[derive(Debug)]
+pub enum SerializationError {
+    /// During serialization, the output buffer was of the wrong size.
+    BufferWrongSize,
+    /// During serialization, we didn't have enough space to write extra info.
+    NotEnoughSpace,
+    /// During serialization, the data was invalid.
+    InvalidData,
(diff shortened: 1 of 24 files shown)


algebra/src/serialize/error.rs
```diff
@@ -0,0 +1,51 @@
...
+#[derive(Debug)]
+pub enum SerializationError {
+    /// During serialization, the output buffer was of the wrong size.
+    BufferWrongSize,
+    /// During serialization, we didn't have enough space to write extra info.
+    NotEnoughSpace,
+    /// During serialization, the data was invalid.
+    InvalidData,
+    /// During serialization, we countered an I/O error.
+    IoError(io::Error),
```
Pratyush, line 17: We don't need this variant I think?

algebra/src/serialize/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        impl<P: $params> CanonicalDeserialize for GroupAffine<P> {
+            fn deserialize(bytes: &[u8], _: &mut [bool]) -> Result<Self, crate::serialize::SerializationError>
+                where Self: Sized {
+                let mut extra_info_buf = [false; 2];
+                let x = P::BaseField::deserialize(bytes, &mut extra_info_buf)?;
+                if extra_info_buf[1] {
+                    return Ok(Self::zero())
+                }
+                GroupAffine::<P>::get_point_from_x(x, extra_info_buf[0])
```
Pratyush: Should we add checks for ensuring the point is in the prime subgroup here?

algebra/src/serialize/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        impl<P: $params> CanonicalDeserialize for GroupAffine<P> {
+            fn deserialize(bytes: &[u8], _: &mut [bool]) -> Result<Self, crate::serialize::SerializationError>
+                where Self: Sized {
+                let mut extra_info_buf = [false; 1];
+                let x = P::BaseField::deserialize(bytes, &mut extra_info_buf)?;
+                if x == P::BaseField::zero() {
+                    return Ok(Self::zero())
+                }
+                GroupAffine::<P>::get_point_from_x(x, extra_info_buf[0])
```
Pratyush: Should we add a prime-subgroup check here?

Pratyush (approved):

algebra/src/serialize/error.rs
kobigurk: The read/write use this variant

Pratyush:
LGTM, I'll let CI run

PR #76, opened 2020-01-27
no_std for algebra and r1cs-core

debris:
Hey! This is an ongoing attempt to make zexe `no_std` compatible. 

- `algebra` compiles with `no_std`
  - removed `byteorder` dependency
  - uses `algebra/src/io.rs` as an `std::io` counterpart
  - uses `algebra/serialize/Error` as `std::error::Error` trait counterpart
  - uses `rand::test_rng()` for generating random numbers in tests
  - split `algebra-benches` into it's own crate - [reason](https://github.com/scipr-lab/zexe/pull/76#issuecomment-579924945)
- `r1cs-core` compiles with `no_std`
- ~`ff-fft` compiles with `no_std`~
  - I did a bunch of changes to `ff-fft`, but eventually decided to postpone `no-std` support for this library till the next PR. This PR is already big enough
- removed floating point operations (#79)
- removed `println!` from non-test code
- added `no-std` bare metal build for `algebra`, `r1cs-core` and ~`ff-fft`~ to `.travis.yml`
- updated `smallvec` to `1.2`, because it contains a fix for [this issue](https://github.com/servo/rust-smallvec/issues/197)


Pratyush:
Hey! Awesome! This would indeed be very useful!

Btw, you don't need to remove byteorder; you can use it with `default-features = false` to have it run in `no_std` mode.

algebra/src/lib.rs
```diff
@@ -33,6 +34,20 @@
 )]
 #![forbid(unsafe_code)]
 
+#[cfg(not(feature = "std"))]
+macro_rules! println {
```
Pratyush: Do we use this macro anywhere in the crate? If we do, it's probably a bug.

algebra/src/lib.rs
```diff
@@ -82,3 +97,5 @@ pub mod prelude {
 
     pub use num_traits::{One, Zero};
 }
+
+mod fake_io;
```
Pratyush: Can we rename this to `io`, and make it conditionally equal the `std` module:
```suggestion
#[cfg(not(feature = "std"))]
mod io;
#[cfg(feature = "std")]
pub use std::io;
```

algebra/src/serialize/error.rs
```diff
@@ -14,26 +15,26 @@ pub enum SerializationError {
...
-impl Error for SerializationError {
-    fn source(&self) -> Option<&(dyn Error + 'static)> {
-        None
-    }
-}
+//impl Error for SerializationError {
+    //fn source(&self) -> Option<&(dyn Error + 'static)> {
+        //None
+    //}
+//}
```
Pratyush: Maybe instead of commenting these out, replace it with a `#[cfg(feature = "std")]`?

algebra/src/serialize/error.rs
```diff
@@ -14,26 +15,26 @@ pub enum SerializationError {
...
 }
 
 impl fmt::Display for SerializationError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        if let SerializationError::IoError(e) = self {
-            write!(f, "I/O error: ")?;
-            e.fmt(f)
+        if let SerializationError::IoError(_) = self {
+            write!(f, "I/O error")
+            //e.fmt(f)
```
Pratyush: we can just delete this line instead of commenting it out.

algebra/src/to_field_vec.rs
```diff
@@ -6,9 +8,10 @@ use crate::{
         ProjectiveCurve,
     },
     Fp2, Fp2Parameters, FpParameters, Field, PrimeField,
+    serialize::SerializationError,
 };
 
-type Error = Box<dyn std::error::Error>;
+type Error = Box<SerializationError>;
```
Pratyush: We can just replace this with `SerializationError` now, right? No need for the `Box`?

algebra/src/bytes.rs
```diff
@@ -282,17 +282,18 @@ impl<'a, T: 'a + ToBytes> ToBytes for &'a T {
...
-    }
-}
+//impl FromBytes for Vec<u8> {
+    //#[inline]
+    //fn read<R: Read>(mut reader: R) -> IoResult<Self> {
+        //let mut buf = Vec::new();
+        //let _ = reader.read_to_end(&mut buf)?;
+        //Ok(buf)
+    //}
+//}
```
Pratyush: We don't need to delete this, right? It should be available even in `no_std` because we're importing `alloc`?

Pratyush (commented):
Added some comments on the changes to `algebra`

algebra/src/lib.rs
debris: Yes, a bunch of tests uses it. Should I replace it with a `log` crate?

algebra/src/lib.rs
debris: yes, I think it's doable

algebra/src/serialize/error.rs
debris: I was thinking of removing it completely, but of course, I can add `#[cfg(feature = "std")]` for backwards compatibility

algebra/src/serialize/error.rs
debris: yes, the pr is not finished yet 😅  I wouldn't commit in commented out code :)

algebra/src/to_field_vec.rs
debris: Sure. I didn't remove the `Box`, cause I thought that it may be here for performance reasons (e.g. to avoid allocating space on stack for a big `Error`), but I don't think it is a case here

Pratyush: Yeah, earlier I didn't want to do the work of defining a custom Error enum, so I took the lazy shortcut =P

algebra/src/lib.rs
debris: done

algebra/src/to_field_vec.rs
debris: done :)

algebra/src/lib.rs
Pratyush: hmm maybe you can add
```rust
#[cfg(and(test, not(feature = "std")))]
#[macro_use]
extern crate std;
```

algebra/src/bytes.rs
debris: I removed it cause it's not used anywhere. I could re-add it, but that would imply that we have to add `read_to_end` to `algebra::io::Read`

debris:
I think the PR is ready. Once merged, I'll work on adding support to the remaining libraries.

> Btw, you don't need to remove byteorder; you can use it with default-features = false to have it run in no_std mode.

Since rust 1.32, `byteorder` is not really needed, cause all unsigned integer primitives have [`from_le_bytes`](https://doc.rust-lang.org/std/primitive.u64.html#method.from_le_bytes) method

kobigurk:
Naive question since I haven't looked closely - does this allow verifying
proofs on wasm?

debris:
@kobigurk not yet. I have another branch that I haven't pushed yet (it's not finished) that allows to do that

debris:
update*
I pushed the branch here: https://github.com/scipr-lab/zexe/compare/master...debris:groth16-prover?expand=1

debris:
I did some tests, and I need to work on this PR a little bit more. It's because of how `cargo` build system works. 

`no-std` version of `algebra` crate depends on `no-std` version of `num-traits`. Unfortunately, one of `dev-dependencies` of `algebra` is `criterion` which depends on `std` version of `num-traits`. Because of that, `cargo`  builds `num-traits` only once - with `std` and links exactly this version of `num-traits` to `algebra`.

this build issue applies also to several other dependencies

I believe the only way to solve this problem is to separate `algebra` and `algebra-benches` from each other

Pratyush:
Yeah the unification of features across `dev-dependencies` is *very* annoying. We can also try submitting a PR to upstream `criterion` to make them use `num-traits` in `no_std` mode

Pratyush:
Ok I made the PR, but it might make sense to have a separate benches crate anyway, just to avoid future fragility.

algebra/Cargo.toml
```diff
@@ -37,4 +36,5 @@ criterion = "0.3.1"
 rand_xorshift = { version = "0.2" }
 
 [features]
+std = []
 parallel = [ "rayon" ]
```
Pratyush:
```suggestion
parallel = [ "rayon", "std" ]
```

groth16/Cargo.toml
```diff
@@ -23,7 +23,7 @@ edition = "2018"
 ################################# Dependencies ################################
 
 [dependencies]
-algebra = { path = "../algebra", features = [ "parallel" ] }
+algebra = { path = "../algebra", features = [ "parallel", "std" ] }
```
Pratyush:
```suggestion
algebra = { path = "../algebra", features = [ "parallel" ] }
```

Pratyush: Though, if you want these crates to also be `no_std`, you probably need to drop the `parallel` feature.

I'd recommend adding a `parallel` feature to this crate, and then making that activate `algebra/parallel`.

Pratyush (changes requested):
Left some comments. Also, to make `ff-fft` `no_std`, you probably need to have a conditional `parallel` feature that activates the parallelization infrastructure https://github.com/scipr-lab/zexe/blob/master/ff-fft/src/multicore.rs

When `parallel` is off the `scope` method would presumably forward to just `chunks` or something. Maybe something similar to https://github.com/matter-labs/bellman/blob/ff783cb207b4d61b8444a4aad91151ee498ccb32/src/singlecore.rs#L56

groth16/Cargo.toml
Pratyush: (same for `gm17`)

debris: I'll work on this in the next PR

debris:
> Left some comments. Also, to make ff-fft no_std, you probably need to have a conditional parallel feature that activates the parallelization infrastructure https://github.com/scipr-lab/zexe/blob/master/ff-fft/src/multicore.rs
> 
> When parallel is off the scope method would presumably forward to just chunks or something. Maybe something similar to https://github.com/matter-labs/bellman/blob/ff783cb207b4d61b8444a4aad91151ee498ccb32/src/singlecore.rs#L56

I decided to postpone making `ff-fft` `no-std` until the next PR. This one is already big enough. And I think it's ready for another review :)

algebra/src/rand.rs
```diff
@@ -16,3 +16,14 @@ where
         rng.sample(Standard)
     }
 }
+
+pub fn test_rng() -> rand::rngs::StdRng {
```
Pratyush, line 21: ~We can put this under `#![cfg(test)]` so that it's not exported.~

I see that it's used in `ff-fft` and other crates. Maybe we can add a line of documentation indicating that this should be used only for testing purposes, and not for any real-world usage?

r1cs-core/src/impl_lc.rs
```diff
@@ -13,14 +13,14 @@ impl<F: Field> AsRef<[(Variable, F)]> for LinearCombination<F> {
...
-        LinearCombination(smallvec![(var, coeff)])
+        LinearCombination(smallvec!((var, coeff)))
     }
 }
 
 impl<F: Field> From<Variable> for LinearCombination<F> {
     #[inline]
     fn from(var: Variable) -> Self {
-        LinearCombination(smallvec![(var, F::one())])
+        LinearCombination(smallvec!((var, F::one())))
```
Pratyush: Could you undo the change from `smallvec![...]`? It's supposed to be used similarly to the `vec![...]` macro

algebra-benches/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+description = "A benchmark library for finite fields and elliptic curves"
+homepage = "https://libzexe.org"
+repository = "https://github.com/scipr/zexe"
+documentation = "https://docs.rs/algebra/"
+keywords = ["cryptography", "finite fields", "elliptic curves", "pairing"]
+categories = ["cryptography"]
+include = ["Cargo.toml", "src", "README.md", "LICENSE-APACHE", "LICENSE-MIT"]
+license = "MIT/Apache-2.0"
+edition = "2018"
+
```
Pratyush, line 22: Can we add a `publish = false` here so that it isn't accidentally published to `crates.io`, and it's clear that it's meant only for local benchmarking?

Pratyush (changes requested):
Awesome! The PR looks mostly good to me. CI is failing because (I think because `r1cs-std` doesn't yet use `r1cs-core`/`algebra` under the `std` flag).

Oh and if possible, could you squash the commits into logically related ones?

debris:
> Awesome! The PR looks mostly good to me. CI is failing because (I think because `r1cs-std` doesn't yet use `r1cs-core`/`algebra` under the `std` flag).

I see... To fix the problem, I decided to make `algebra` and `r1cs-core` `std` by default. It's an approach that many other projects also have (e.g. `serde`). To compile it for `no-std`, one needs to set `default-features = false`

> Oh and if possible, could you squash the commits into logically related ones?

Hmm... is it possible to just squash all of them into a single commit using github's `Squash and merge` option? 😅

debris:
CI is green 🎉

Pratyush:
Awesome, merged! Thank you so much for the PR!

PR #80, opened 2020-01-28
More light dependency tweaking

huitseeker:


crypto-primitives/Cargo.toml
```diff
@@ -26,7 +26,6 @@ algebra = { path = "../algebra" }
 r1cs-core = { path = "../r1cs-core", optional = true }
 r1cs-std = { path = "../r1cs-std", optional = true }
 gm17 = { path = "../gm17", optional = true }
-groth16 = { path = "../groth16", optional = true }
```
Pratyush: Hmm we use it in the crate conditionally: https://github.com/scipr-lab/zexe/blob/ee885534fcd0d64eda0a356fb7929ccbc770ffcc/crypto-primitives/src/nizk/mod.rs#L9

huitseeker: :man_facepalming:

PR #85, opened 2020-01-31
no-std for ff-fft and groth16

debris:
built on top of pr #76

### DIFF
--- a/ff-fft/src/polynomial/dense.rs
+++ b/ff-fft/src/polynomial/dense.rs
@@ -158,15 +158,9 @@ impl<F: PrimeField> DensePolynomial<F> {
     pub fn mul_by_vanishing_poly(&self, domain: EvaluationDomain<F>) -> DensePolynomial<F> {
         let mut shifted = vec![F::zero(); domain.size()];
         shifted.extend_from_slice(&self.coeffs);
-
-        #[cfg(feature = "parallel")]
-        let shifted_iter = shifted.par_iter_mut();
-
-        #[cfg(not(feature = "parallel"))]
-        let shifted_iter = shifted.iter_mut();
-
-        shifted_iter.zip(&self.coeffs).for_each(|(s, c)| *s -= c);
-
+        cfg_iter_mut!(shifted)
+            .zip(&self.coeffs)
+            .for_each(|(s, c)| *s -= c);
         DensePolynomial::from_coefficients_vec(shifted)
     }
 
--- a/ff-fft/src/polynomial/dense.rs
+++ b/ff-fft/src/polynomial/dense.rs
@@ -9,6 +9,8 @@ use core::{
 use crate::{DenseOrSparsePolynomial, EvaluationDomain, Evaluations};
 use algebra::{Field, PrimeField};
 use rand::Rng;
+
+#[cfg(feature = "parallel")]
 use rayon::prelude::*;
 
 /// Stores a polynomial in coefficient form.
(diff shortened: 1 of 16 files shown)


ff-fft/src/domain_seq.rs
```diff
@@ -0,0 +1,422 @@
+//! This module contains an `EvaluationDomain` abstraction for
```
Pratyush: Do we need this file? I think we can just `#cfg[feature = "std"]` out anything that depends on `rayon`.

ff-fft/src/polynomial/dense.rs
```diff
@@ -104,11 +106,24 @@ impl<F: Field> DensePolynomial<F> {
         }
         assert_eq!(powers_of_point.len(), self.coeffs.len());
         let zero = F::zero();
-        powers_of_point
```
Pratyush, line 107: Let's do something similar to here: https://github.com/scipr-lab/zexe/blob/8a9343ad03e24cf278598e19c9a7b928b3ab729e/algebra/src/msm/fixed_base.rs#L80

This avoid duplicating the rest of the iterator chain.

ff-fft/src/polynomial/dense.rs
```diff
@@ -143,10 +158,22 @@ impl<F: PrimeField> DensePolynomial<F> {
     pub fn mul_by_vanishing_poly(&self, domain: EvaluationDomain<F>) -> DensePolynomial<F> {
         let mut shifted = vec![F::zero(); domain.size()];
         shifted.extend_from_slice(&self.coeffs);
-        shifted
```
Pratyush, line 146: Let's do something similar to here: https://github.com/scipr-lab/zexe/blob/8a9343ad03e24cf278598e19c9a7b928b3ab729e/algebra/src/msm/fixed_base.rs#L80

This avoid duplicating the rest of the iterator chain.

Pratyush (changes requested):
Cool thanks! I merged #76 and also added a new PR (#86) that simplifies the code in `domain.rs`, so if you could rebase that'd be great.

ff-fft/src/polynomial/dense.rs
debris: fixed

ff-fft/src/polynomial/dense.rs
debris: in this case the iterator chain is not duplicated, cause `fold/reduce` function from `rayon` has a different `signature` than `fold` from `std` and it requires us to pass `Fn` as accumulator

[rayon-docs#method.fold](https://docs.rs/rayon/1.3.0/rayon/iter/trait.ParallelIterator.html#method.fold)

[std-docs#method.fold](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold)

ff-fft/src/domain_seq.rs
debris: you are right, after simplifying the code in #86 it is much easier to just add `#cfg[feature = "std"]` everywhere it is needed

Pratyush: To clarify, I meant that we can use `#cfg[feature = "std"]` on the original `domain.rs`.

debris:
I think it's ready for another review :)

Pratyush:
Hmm it looks generally good. I noticed that there's a lot of code duplication due to the `cfg`s. Maybe we can define a macro like `cfg_iter` that does something like:
```rust
macro_rules! cfg_iter {
    ($e:expr) => {
        #[cfg(feature = "parallel")]
		let result = $e.par_iter()
		#[cfg(not(feature = "parallel"))]
		let result = $e.iter();
		result
    }
}
```
This can then be used something like ``` let result = cfg_iter!(array_to_be_iterated_over).zip(other_array)...```.

debris:
Added 3 new macros `cfg_iter`, `cfg_iter_mut` and `cfg_into_iter` :)

Pratyush:
Awesome! This is great! I’ll merge once CI passes :)

PR #92, opened 2020-02-05
r1cs-std works with no-std

debris:
changes:
- `test-constraint-system` uses `BTreeMap` instead of `radix_trie`
- `r1cs-std` supports `no-std`


Pratyush:
Hmm for the radix-trie change you can just replace it with a standard map type, like BTreeMap. (I'd prefer to avoid splitting it out into a separate crate.)

r1cs-std/src/test_constraint_system.rs
```diff
@@ -63,6 +62,7 @@ impl<ConstraintF: Field> TestConstraintSystem<ConstraintF> {
         }
     }
 
+    #[allow(unused_variables)]
```
Pratyush, line 65: Why is this needed here?

debris: a redefinition of `println` on `no-std` doesn't do anything

Pratyush: ahh so the idea is that since the no_std version of the macro doesn't do anything with the variable, it leads to the variable being left unused. Ok, makes sense.

PR #93, opened 2020-02-05
gm17 works with no-std

debris:


gm17/Cargo.toml
```diff
@@ -22,19 +22,21 @@ edition = "2018"
...
 [dev-dependencies]
 csv = { version = "1" }
 
 [features]
+default = ["parallel"]
+std = ["algebra/std", "ff-fft/std", "r1cs-core/std"]
+parallel = ["std", "algebra/parallel", "ff-fft/parallel", "rayon"]
 print-trace = [ "bench-utils/print-trace" ]
 
 [[example]]
```
Pratyush, line 42: I think you have to add `required-features = ["std"]` like [here](https://github.com/scipr-lab/zexe/blob/ac165844ab6fa6fc7fef88cf3f6a60e810d97fb3/groth16/Cargo.toml#L46)

PR #96, opened 2020-02-11
crypto-primitives no-std

debris:
major changes:

- `crypto-primitives` works with `no-std`
- moved `crypto-primitives/benches` to a new crate `cp-benches`, because of `criterion` dependency


crypto-primitives/src/merkle_tree/mod.rs
```diff
@@ -255,23 +255,27 @@ impl std::fmt::Display for MerkleTreeError {
...
 fn log2(number: usize) -> usize {
-    (number as f64).log2() as usize
+    algebra::log2(number) as usize
 }
 
 /// Returns the height of the tree, given the size of the tree.
 #[inline]
 fn tree_height(tree_size: usize) -> usize {
-    log2(tree_size + 1)
+    log2(tree_size)
```
debris, line 278: `algebra::log2` returns approximate value of `log2`. To make tests pass, I had to remove `+1` from here. I am not sure whether this changes the logic significantly

Pratyush: Maybe we can use this: https://users.rust-lang.org/t/logarithm-of-integers/8506/5

Pratyush: Nvm I should have read the code before commenting 🤦🏾‍♂️

crypto-primitives/src/crh/bowe_hopwood/mod.rs
```diff
@@ -149,6 +152,32 @@ impl<G: Group, W: PedersenWindow> FixedLengthCRH for BoweHopwoodPedersenCRH<G, W
                     .reduce(G::zero, |a, b| a + &b)
             })
             .reduce(G::zero, |a, b| a + &b);
+
```
Pratyush: You can avoid duplication by doing `.sum()`; the latest master adds in support for this :)

crypto-primitives/src/crh/pedersen/mod.rs
```diff
@@ -112,6 +114,22 @@ impl<G: Group, W: PedersenWindow> FixedLengthCRH for PedersenCRH<G, W> {
                 encoded
             })
             .reduce(G::zero, |a, b| a + &b);
+
+        #[cfg(not(feature = "parallel"))]
```
Pratyush: Same here

Pratyush (changes requested):
Just the small changes to avoid duplication, and we should be good to go! Thanks for the PR!

debris:
ready for another review :)

Pratyush:
Looks good, thanks!

PR #99, opened 2020-02-17
Serialize and Deserialize Uncompressed

gakonst:
This PR defines a new serialization trait which directly serializes a group element in its uncompressed form. In addition, I found it reasonable to add the serialization trait bounds in the `AffineCurve` trait.

### DIFF
--- a/algebra/src/serialize/mod.rs
+++ b/algebra/src/serialize/mod.rs
@@ -7,11 +7,22 @@ pub trait CanonicalSerialize {
         extra_info: &[bool],
         output_buf: &mut [u8],
     ) -> Result<(), SerializationError>;
+
     fn buffer_size() -> usize;
+}
+
+pub trait GroupSerialize: CanonicalSerialize {
+    fn serialize_uncompressed(&self, output_buf: &mut [u8]) -> Result<(), SerializationError>;
 }
 
 pub trait CanonicalDeserialize: CanonicalSerialize {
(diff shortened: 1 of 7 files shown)


algebra/src/serialize/mod.rs
```diff
@@ -144,6 +155,34 @@ macro_rules! impl_sw_curve_serializer {
                 Ok(p)
             }
         }
+
+        impl<P: $params> GroupSerialize for GroupAffine<P> {
+            fn serialize_uncompressed(
+                &self,
+                output_buf: &mut [u8],
+            ) -> Result<(), crate::serialize::SerializationError> {
+                let len = <Self as CanonicalSerialize>::buffer_size();
```
kobigurk: What about using the field buffer size here instead?

kobigurk:
Also seems like it needs `cargo fmt`

algebra/src/serialize/mod.rs
gakonst: Good idea, fixed: https://github.com/scipr-lab/zexe/pull/99/commits/3039fc1d5885753ea4317914744b111d614d2afc#diff-40da71ffac4550ada607d2de65a09100R164

Pratyush:
What do you think about adding these as new methods on the `Canonical{Serialize,Deserialize}` traits? Instead of adding new traits?

gakonst:
This would mean we'd need to implement the uncompressed methods for Fields as well which I wanted to avoid since I did not need it anywhere (and did not see when it'd be used). When would (de)serializing uncompressed fields be useful?

PR #102, opened 2020-02-25
Add a pre-commit hook for rustfmt

huitseeker:
The commit hook is set up (permanently) on the first build of `bench-utils`, through a build-script of that package.

The pre-commit hook will reformat files (& report those it had to modify) before a commit.
This is to alleviate the problems I've seen running rustfmt lately.


Pratyush:
Hmm I'd prefer not to modify the build-script. Perhaps we can instead add a separate bash script that can be manually invoked to set up the git hooks?

Another alternative is to add a new entry to the travis matrix that just checks that the formatting is correct.

huitseeker:
@Pratyush we already have a portion in CI that checks the formatting is correct, but it tripped you, @howardwu  and @gakonst within the last week.

Rather than modify the bench-utils build, I ~~can add~~ added an empty 'scripts' package at the end of the workspace which goal is to set up the git property pointing to the hook ~~if that's more amenable to you~~.

Pratyush:
Hmm do you think it makes sense to just add a `CONTRIBUTING.md` which has guidelines for contributing code (and would include running the script as one of the guidelines). We can adapt something like https://github.com/servo/servo/blob/master/CONTRIBUTING.md as a starting point?

Basically, the reason why I want to avoid adding another crate is because adding the hook is usually a one-time change for regular contributors, while for one-off contributors/small PRs the script crate might never be compiled, leading to the same push-CI-format dance. It also adds another crate to the fairly crowded repo.

We can also disable the format check in CI if you wish; I underestimated the trouble it causes.

huitseeker:
@Pratyush I replaced the installation of the hook with a simple bash script

Pratyush:
Perfect, thanks!

PR #108, opened 2020-02-27
feat(groth16/gm17): make KeyPairAssembly fields public

gakonst:
As title.

### DIFF
--- a/gm17/src/generator.rs
+++ b/gm17/src/generator.rs
@@ -34,14 +34,14 @@ where
 }
 
 /// This is our assembly structure that we'll use to synthesize the
-/// circuit into a SAP.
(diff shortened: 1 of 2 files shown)


Pratyush:
I think it's fine to make the fields `pub`; I have a refactor in mind that (as a side effect) would enable this anyway.

gm17/src/generator.rs
```diff
@@ -34,14 +34,14 @@ where
 }
 
 /// This is our assembly structure that we'll use to synthesize the
-/// circuit into a SAP.
+/// circuit into a QAP.
```
gakonst:
```suggestion
/// circuit into a SAP.
```

PR #110, opened 2020-02-28
Add back static bounds on `Field`, `Group`, and `{Projective,Affine}Curve`

Pratyush:


Pratyush:
See here for why: https://stackoverflow.com/a/40053649

PR #114, opened 2020-02-28
Add exponentiation by constant to FieldGadget

ValarDragon:


Pratyush:
Could you slightly generalize the implementation to be like the corresponding code in Field:
https://github.com/scipr-lab/zexe/blob/4ae139cc80a7763e4e5999e36cdfdc4f57148b9d/algebra-core/src/fields/mod.rs#L118

r1cs-std/src/fields/mod.rs
```diff
@@ -431,9 +431,10 @@ pub(crate) mod tests {
                 .unwrap()
         );
 
+        // a * a * a = a^3
         assert_eq!(
             a_native * &(a_native * &a_native),
-            a.pow_by_constant(cs.ns(|| "test_pow"), 3)
+            a.pow_by_constant(cs.ns(|| "test_pow"), &[3])
```
Pratyush: The test is currently failing because of the repeated namespacing
```suggestion
            a.pow_by_constant(cs.ns(|| "test_constant_pow"), &[3])
```

ValarDragon: Oh woops, not quite sure why it passed on my machine.

ValarDragon: As a side note, is it important that this errors? It feels like something that should cause a warning at most.

Pratyush: After our discussion I'm going to make namespace generation (mostly) automated, so this should not be a problem after that =)

r1cs-std/src/fields/mod.rs
```diff
@@ -224,6 +224,31 @@ pub trait FieldGadget<F: Field, ConstraintF: Field>:
...
+            {
+                res.square_in_place(cs.ns(|| format!("square for bit {:?}", i)))?;
+            }
+    
+            if !i
+            {
+                found_one = true;
+                res.mul_in_place(cs.ns(|| format!("mul for bit {:?}", i)), self)?;
+            }
+        }
```
Pratyush:
```suggestion
        for (i, bit) in BitIterator::new(exp).enumerate() {
            if found_one {
                res.square_in_place(cs.ns(|| format!("square for bit {:?}", i)))?;
            }
    
            if !bit {
                found_one = true;
                res.mul_in_place(cs.ns(|| format!("mul for bit {:?}", i)), self)?;
            }
        }
```

Pratyush:
btw the tests are failing with incorrect output https://travis-ci.org/scipr-lab/zexe/jobs/656522556

ValarDragon:
I'm pretty confused, am I testing wrong?
```
z/r1cs-std> git branch
  master
* pow
z/r1cs-std> cargo test
Finished test [unoptimized + debuginfo] target(s) in 0.04s
<snip>
test result: ok. 26 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

EDIT: If I run the same command in CI from `/zexe`, I can reproduce. Its odd that `cargo test` from `/zexe` and from `r1cs-std` didn't catch this.

Pratyush:
I think you need to try it with cargo test —all-features

ValarDragon:
Thanks! Its fixed, the error got introduced when I switched to bit iterator syntax.

Feels a bit weird that the default for tests on r1cs-std is to have the r1cs tests off though

Pratyush:
They test the constraints, but the curve and field gadgets are behind feature-flags because not everyone might want all the different options

Pratyush:
(btw thanks for the PR!)

ValarDragon:
Thanks for the expl!

PR #115, opened 2020-02-28
chore(groth16): remove redundant elements from Parameters

gakonst:
As title, these elements were already part of the VK


kobigurk (commented):
Confirmed, we've chatted about it :)

PR #117, opened 2020-02-28
Generalize FFT infrastructure to work with group elements as well

Pratyush:
cc @gakonst @kobigurk I think this is what you need for the MPC, right?

### DIFF
--- a/ff-fft/src/test.rs
+++ b/ff-fft/src/test.rs
@@ -1,22 +1,28 @@
 use crate::domain::*;
-use algebra::bls12_381::Bls12_381;
-use algebra_core::{test_rng, PairingEngine, UniformRand};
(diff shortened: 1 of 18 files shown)


ff-fft/src/test.rs
```diff
@@ -37,7 +43,8 @@ fn fft_composition() {
 
     let rng = &mut test_rng();
 
-    test_fft_composition::<Bls12_381, _>(rng);
+    test_fft_composition::<Fr, Fr, _>(rng);
+    test_fft_composition::<Fr, G1Projective, _>(rng);
```
Pratyush, line 47: See here for an example of using FFTs with group elements

kobigurk:
Thanks @Pratyush! I believe this should work for us.

Issue #125, opened 2020-03-04
Serialization Derive Macro does not support Tuples


gakonst:
As title, ref #104. Example:

```rust
#[derive(Debug, CanonicalSerialize)]
struct TestCase<E: PairingEngine> {
    a: (E::G1Affine, E::G1Affine), // this makes the macro fail
    b: E::G1Affine, // this works
    c: Vec<E::G2Affine>, // this works
}
```

It'd be nice if it would just serialize `a.0` followed by `a.1`

Sidenote: In order to use the macro, I had to explicitly import `algebra-core` to my package. It'd be nice if the macro were usable via importing just `algebra` as well.

i.e. go from:
```
zexe_algebra = { package = "algebra", git = "https://github.com/scipr-lab/zexe", version = "0.1.0" }
zexe_algebra_core_derive = { package = "algebra-core-derive", git = "https://github.com/scipr-lab/zexe", version = "0.1.0" }
algebra_core = { package = "algebra-core", git = "https://github.com/scipr-lab/zexe", version = "0.1.0" }
```

to
`zexe_algebra = { package = "algebra", git = "https://github.com/scipr-lab/zexe", version = "0.1.0", features = ["derive"] }`

cc @paberr

paberr:
I'll take a look into it. :)

paberr:
~I opened a PR with the functionality you were looking for.~
Ok, I just noticed that re-exporting it in both algebra and algebra-core results in problems when compiling the tests.

paberr:
So, the problem with not having `algebra-core` as a dependency is:
- The derive macro generates code that refers to types like `::algebra_core::SerializationError`.
- Since `algebra` also re-exports these types, I tried choosing between `algebra_core` and `algebra` using a feature flag at compile time.
- However, since feature flags are active within the whole workspace/project, the following will happen:
    1. Let's suppose you write your own crate that only depends on `algebra` and `gm17`.
    2. You want to use the derive macro using `algebra` only. Then, the feature will be set for the whole workspace. The derive macros will thus use `::algebra::<...>` as a path.
    3. However, gm17/groth16 only depend on `algebra-core`. This will cause `gm17` to not compile anymore.

To avoid this mess, I guess the best way is to still require `algebra-core` as a dependency.
You don't need to explicitly depend on the `algebra-core-derive` anymore, but can use a feature flag in `algebra-core`.

If someone has another idea how to realise this, I'm open for suggestions.

gakonst:
[...]

PR #126, opened 2020-03-04
Add support for tuples in (de-)serialization derive macro

paberr:
Additionally allows re-export of (de-)serialization derive macro from algebra-core via feature flag.

This achieves what was requested in #125.
In particular, it allows to derive (de-)serialization for nested tuples as shown in the following example:
```rust
#[derive(Debug, CanonicalSerialize, CanonicalDeserialize)]
struct TestCase<E: PairingEngine> {
    a: (E::G1Affine, (E::G1Affine, E::G2Affine)), // this works
    b: E::G1Affine,                               // this works
    c: Vec<E::G2Affine>,                          // this works
}
```

The derive functionality can now be used by depending on:
```
algebra_core = { ..., features = [ "derive" ] }
```

I also tried making it available via `algebra` only, but this results in several problems as described in #125 .

### DIFF
--- a/algebra-core/Cargo.toml
+++ b/algebra-core/Cargo.toml
@@ -22,6 +22,7 @@ edition = "2018"
 ################################# Dependencies ################################
 
 [dependencies]
+algebra-core-derive = { path = "algebra-core-derive", optional = true }
 derivative = { version = "1", features = ["use_core"] }
 num-traits = { version = "0.2", default-features = false }
 rand = { version = "0.7", default-features = false }
@@ -31,6 +32,7 @@ rayon = { version = "1", optional = true }
 rand_xorshift = "0.2"
 
 [features]
-default = ["std"]
+default = [ "std" ]
 std = []
-parallel = ["std", "rayon"]
+parallel = [ "std", "rayon" ]
+derive = [ "algebra-core-derive" ]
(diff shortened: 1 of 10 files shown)


Pratyush:
Does this mean that a crate that imports `algebra` but wants to use the derives would need to import `algebra-core` too? Is it possible to also add a `derive` feature to `algebra/Cargo.toml` that activates `algebra-core/derive`?

paberr:
> Does this mean that a crate that imports `algebra` but wants to use the derives would need to import `algebra-core` too?

Sadly, yes. That's also the case right now.
The reason for this is that the derived code needs to refer types like `Write` or `SerializationError` and does so via `::algebra_core::SerializationError`.
> Is it possible to also add a `derive` feature to `algebra/Cargo.toml` that activates `algebra-core/derive`?

I tried to do that, but it isn't straightforward. Adding the derive feature is easy, but if you try using it without having `algebra-core` as a dependency, the `::algebra_core::SerializationError` will cause a compilation error.

I also tried choosing between `::algebra_core` and `::algebra` in the derived code based on whether your dependency is `algebra-core` or `algebra`. The resulting problems are described [here](https://github.com/scipr-lab/zexe/issues/125#issuecomment-594503350).

One way forward would be not to specify the full path in the derived code and rely on the user of the derive macro to import all relevant types. This could be done by something like:
```rust
use algebra::serialization::*;

#[derive(CanonicalSerialize, CanonicalDeserialize)]
struct Test {
    <...>
}
```

Pratyush:
Yeah let's go for the latter option, it seems more flexible.

(I think it's what `serde` uses too!)

paberr:
I added the feature to `algebra`, rebased the branch on the current master and added examples for the trait derivation.

Pratyush:
OK this looks good to me. Thanks for the PR!

PR #127, opened 2020-03-05
added UInt64 and sparse tree

weikengchen:
First time commit. It passes the tests I wrote.

Please give some code review.


weikengchen:
Closed. It seems that UInt64 needs to catch up with the library change.

PR #128, opened 2020-03-05
added UInt64 and sparse tree

weikengchen:
First time commit. It passes the tests I wrote.

Please give some code review.

### DIFF
--- a/r1cs-std/src/bits/uint64.rs
+++ b/r1cs-std/src/bits/uint64.rs
@@ -5,7 +5,7 @@ use r1cs_core::{ConstraintSystem, LinearCombination, SynthesisError};
 use crate::{
     boolean::{AllocatedBit, Boolean},
     prelude::*,
-    Assignment,
+    Assignment, Vec,
 };
 
 /// Represents an interpretation of 64 `Boolean` objects as an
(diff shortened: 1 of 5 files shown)


ValarDragon:
The Uint64 impl is independent of the Sparse Merkle Tree right? 

If so, perhaps each can be split into independent PRs for ease of reviewing?

r1cs-std/src/bits/uint64.rs
```diff
@@ -0,0 +1,544 @@
...
+    /// Perform modular addition of several `UInt64` objects.
+    pub fn addmany<ConstraintF, CS>(mut cs: CS, operands: &[Self]) -> Result<Self, SynthesisError>
+    where
+        ConstraintF: PrimeField,
+        CS: ConstraintSystem<ConstraintF>,
+    {
+        // Make some arbitrary bounds for ourselves to avoid overflows
+        // in the scalar field
+        assert!(ConstraintF::Params::MODULUS_BITS >= 128);
+        assert!(operands.len() >= 2); // Weird trivial cases that should never happen
```
ValarDragon, line 170: Perhaps if operands.len() == 1, make it return operands[0]? 

Otherwise thats an (imo) reasonable edge-case the caller may have to add a try-catch for.

weikengchen:
Closed. I will separate the UInt64 and the sparse Merkle tree one. 

I still need to get rid of the use of the std library, which seems to be abandoned in the recent refactoring.

weikengchen:
@ValarDragon  Any idea on replacement of std::collections::{HashMap, HashSet}?

Pratyush:
You can use `alloc::collections::{HashMap, HashSet}`

weikengchen:
Great!

PR #129, opened 2020-03-05
Add UInt64 used for representing Merkle tree node locations

weikengchen:
Added UInt64, which resembles UInt32. It is planned to use UInt64 later to represent Merkle tree node locations, which will be used in a later PR.

As discussed in https://github.com/scipr-lab/zexe/pull/128, it is useful when `operands.len() == 1`, `addmany` simply returns that operand. (no test has been written for this case)


weikengchen:
Hooray! It passes the check. (Just realized that there is a format check).

Pratyush:
great, thanks!

PR #134, opened 2020-03-07
Merkle sparse tree

weikengchen:
Looking for some general suggestions!

Here is an initial attempt for the Merkle sparse tree, which differs from the current Merkle tree from the following aspects:
- Instead of using a padding tree, it uses a sparse tree, which allows writing to locations in a non-incremental way. 
- Support verifiable update.
- Support verification of tree paths with index checking.

It has a different interface from the original Merkle tree's in that its `new` function does not take a generic array of leaves `&[L]` but rather a `&HashMap<usize, L>`.

This PR is to look for general advice. It is likely that this PR will not be merged into Zexe at this moment because the constraints have not been peer-reviewed and the code surely needs more work.

Any advice on making the Merkle tree's `new` polymorphism to `&[L]` and `&HashMap<usize, L>`? This will allow the sparse Merkle tree to serve as an in-place replacement for the current Merkle tree implementation.

### DIFF
--- a/crypto-primitives/src/lib.rs
+++ b/crypto-primitives/src/lib.rs
@@ -17,6 +17,7 @@ pub(crate) use std::{borrow::ToOwned, boxed::Box, vec::Vec};
 
 pub mod commitment;
 pub mod crh;
+pub mod merkle_sparse_tree;
 pub mod merkle_tree;
 pub mod nizk;
 pub mod prf;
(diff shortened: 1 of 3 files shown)


weikengchen:
turned out that HashMap and HashSet do not exist in `alloc.collections` :)

Instead, it is `BTreeMap` and `BTreeSet`. I will change this.

weikengchen:
Closed.

Pratyush:
Sorry I didn't get a chance to review this in detail! You can leave this open, I'll try to get to it soon!

weikengchen:
It has been changed a lot afterward. I will push a new one when it becomes stable.

PR #136, opened 2020-03-12
Rename arguments to CondSelectGadget for code clarity

ValarDragon:
I think this rename makes it obvious what the order of arguments should be, without resorting to looking at the comment. (I initially expected the arguments to be reversed)


Pratyush:
Cool, thanks! I'll merge once CI passes.

PR #138, opened 2020-03-13
Make Groth16 efficient when ZK isn't needed

### LINKED ISSUE #120: Implement an efficient zero-knowledge-less variant of Groth16
When Groth16 is used for arguments without the need zero-knowledge, it's possible to save some calculations.

Specifically, as can be seen below, the term `rB` in `C` is not needed when `r,s=0` (the random elements required for zero-knowledge). This means we don't need to calculate `[B]_1`, saving a large multiexp. It's still sound, as the verifier algorithm is the same.

image

An example of such an implementation can be seen in https://github.com/filecoin-project/bellman/pull/12/files.

Thanks to @arielgabizon for showing this to me a while ago!

kobigurk:
* Don't compute g_b1 if r is zero
* Set d1,d2,d3 to always be zero - they're not needed for Groth16

Adds a helper function `create_proof_no_zk`.

Closes #120.

### DIFF
--- a/groth16/src/prover.rs
+++ b/groth16/src/prover.rs
@@ -194,6 +194,17 @@ where
     create_proof::<E, C>(circuit, params, r, s)
 }
 
+pub fn create_proof_no_zk<E, C>(
+    circuit: C,
+    params: &Parameters<E>,
+) -> Result<Proof<E>, SynthesisError>
+where
+    E: PairingEngine,
+    C: ConstraintSynthesizer<E::Fr>,
+{
+    create_proof::<E, C>(circuit, params, E::Fr::zero(), E::Fr::zero())
+}
+
 pub fn create_proof<E, C>(
     circuit: C,
     params: &Parameters<E>,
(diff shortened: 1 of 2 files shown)


gakonst:
Checked the PR, LGTM - @Pratyush. Does the proof being deterministic for no zk provide any improvements in other areas?

groth16/src/prover.rs
```diff
@@ -234,7 +239,8 @@ where
     end_timer!(synthesis_time);
 
     let witness_map_time = start_timer!(|| "R1CS to QAP witness map");
-    let (full_input_assignment, h, _) = R1CStoQAP::witness_map::<E>(&prover, &d1, &d2, &d3)?;
+    let (full_input_assignment, h, _) =
```
Pratyush: if d1, d2 and d3 are always zero, why do we need to pass them in here?

kobigurk: This is a witness_map that also works for PGHR13, where d_1, d_2 and d_3 are used for zk. I kept it for the possible future where someone wants that, but maybe we can remove it :) Wdyt?

Pratyush: I think we can remove it, I feel like QAP based SNARKs are going out of fashion haha

kobigurk: I think it was @gakonst who gave a presentation about "Groth16 is not dead yet". Maybe PGHR13 is though :P I'll remove it.

kobigurk: OK, done!

groth16/src/prover.rs
```diff
@@ -240,7 +240,7 @@ where
 
     let witness_map_time = start_timer!(|| "R1CS to QAP witness map");
     let (full_input_assignment, h, _) =
-        R1CStoQAP::witness_map::<E>(&prover, &E::Fr::zero(), &E::Fr::zero(), &E::Fr::zero())?;
+        R1CStoQAP::witness_map::<E>(&prover)?;
```
Pratyush: fmt is failing because this should now be on the previous line =/

kobigurk: Oops...

Pratyush:
Thanks @kobigurk!

PR #143, opened 2020-03-16
Adds a smaller than gadget

kobigurk:
The smaller than gadget works for any pair of numbers that are smaller than (p-1)/2.

There are a few versions available:
* smaller_than - assumes that the numbers are checked to be within that range by the caller.
* smaller_than_strict - does the check internally.
* smaller_than_or_equal_to, smaller_than_or_equal_to_strict

Based on an idea @stefandeml showed me in the past.


Pratyush:
Awesome, thanks Kobi! Quick question: is it possible to generalize this to a PartialOrd like API?

kobigurk:
> Awesome, thanks Kobi! Quick question: is it possible to generalize this to a PartialOrd like API?

Great idea!

r1cs-std/src/smaller_than.rs
```diff
@@ -0,0 +1,250 @@
+use crate::{
+    boolean::Boolean,
+    fields::{fp::FpGadget, FieldGadget},
+    ToBitsGadget,
+};
+use algebra::PrimeField;
+use r1cs_core::{ConstraintSystem, SynthesisError};
+use std::marker::PhantomData;
```
Pratyush:
```suggestion
use core::marker::PhantomData;
```

Pratyush (approved):
@kobigurk we can merge this as is, and let's file an issue for moving to the `PartialOrd` like API

kobigurk:
I want to change the naming of functions to be "safe by default", as you did for the to_bits APIs. Might as well do the PartialOrd change on the way.

Pratyush:
Ok great, didn't want to hold up the PR, but if you're fine with that then there's no rush!

kobigurk:
I've made some changes to the API to appear more cmp-like.
I also made an attempt to make it act like EqGadget, being a trait, but couldn't find a nice way to do it, mostly because it only works for FpGadget at the moment. Any clever ideas? :)

Pratyush:
If the intention is just to use it with `FpGadget`, you can implement these as inherent methods on `FpGadget`; that way you don't have to create a separate struct.

Pratyush:
(Also, it seems that the PR is currently failing because it's using the old `to_bits` API: https://travis-ci.org/github/scipr-lab/zexe/jobs/666033432#L319)

kobigurk:
Alright, let's try this one. Looks nicer, I think.

r1cs-std/src/fields/fp/cmp.rs
```diff
@@ -0,0 +1,344 @@
...
+   
+}
+
+pub struct CmpGadget<ConstraintF: PrimeField> {
+    constraint_field_type: PhantomData<ConstraintF>,
+}
+
+impl<ConstraintF: PrimeField> CmpGadget<ConstraintF> {
+}
+
```
Pratyush:
```suggestion
```

Pratyush:
Great, could you also add some documentation on what each function does? Currently not sure what the difference between some of the functions is.

kobigurk:
Thanks for catching that 🤓Added some docs now.

kobigurk:
@Pratyush let me know what you think :)

Pratyush:
LGTM!

PR #145, opened 2020-03-17
Use self / method syntax for groupmap

imeckler:


imeckler:
So sorry @Pratyush ...

PR #146, opened 2020-03-19
Change default `to_bits` and `to_bytes` functions to safe versions

Pratyush:
Changes the `ToBytesGadget` and `ToBitsGadget` to return maximally constrained output by default. For example, if `f` is a `FieldGadget<Fr>, `f.to_bits(cs)` checks that that the output bits are the bitwise decomposition of the unique representative of `f` that is less than `r`, whereas earlier it returned *any* representative that equaled the value of `f` modulo `r`.

Thanks to @ValarDragon for pointing out how the old default could mislead users and lead to security bugs in constraint code.

### DIFF
--- a/crypto-primitives/src/nizk/gm17/constraints.rs
+++ b/crypto-primitives/src/nizk/gm17/constraints.rs
@@ -388,13 +388,6 @@ where
         }
         Ok(bytes)
     }
-
-    fn to_bytes_strict<CS: ConstraintSystem<ConstraintF>>(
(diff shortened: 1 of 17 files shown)


Pratyush:
cc @kobigurk @howardwu

kobigurk:
That's a great change for developer experience!
I actually needed the non-strict behaviour (since the bits in my case were already verified by the fact they're hashed), but I can see how it was confusing before.

PR #149, opened 2020-03-20
Add constant serialized size trait

paberr:
As per wish of @gakonst, here is a convenience trait that allows to get the serialized size for constant sized elements without requiring an instance of the element.

This also includes a commit that adds proper derivation of the uncompressed variant.
So far, derived serialization always used the compressed variant.


Pratyush:
If the output is always constant, maybe it makes sense to make it an associated const?

paberr:
> If the output is always constant, maybe it makes sense to make it an associated const?

Good idea! I changed it and also added proper derivation of the uncompressed variants.
So far, the derivation didn't implement the uncompressed variant at all (falling back to the compressed).

Pratyush:
Awesome, thanks!

PR #151, opened 2020-03-23
Constraint Counter

gakonst:
Since `TestConstraintSystem` does not let us inspect the number of constraints for circuits without populating them, I thought this could be a nice addition for testing/examining larger circuits.

### DIFF
--- a/r1cs-std/src/test_constraint_counter.rs
+++ b/r1cs-std/src/test_constraint_counter.rs
@@ -17,6 +17,10 @@ impl TestConstraintCounter {
             num_constraints: 0,
         }
     }
+
+    pub fn num_constraints(&self) -> usize {
+        self.num_constraints
+    }
 }
 
 impl<ConstraintF: Field> ConstraintSystem<ConstraintF> for TestConstraintCounter {
--- a/r1cs-std/src/test_constraint_counter.rs
+++ b/r1cs-std/src/test_constraint_counter.rs
@@ -3,15 +3,15 @@ use algebra::Field;
 use r1cs_core::{ConstraintSystem, Index, LinearCombination, SynthesisError, Variable};
 
 /// Constraint counter for testing purposes.
-pub struct TestConstraintCounter {
+pub struct ConstraintCounter {
     pub num_inputs: usize,
     pub num_aux: usize,
     pub num_constraints: usize,
 }
 
-impl TestConstraintCounter {
-    pub fn new() -> TestConstraintCounter {
-        TestConstraintCounter {
+impl ConstraintCounter {
+    pub fn new() -> Self {
+        Self {
             num_aux: 0,
             num_inputs: 0,
             num_constraints: 0,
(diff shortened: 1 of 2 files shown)


r1cs-std/src/test_constraint_counter.rs
```diff
@@ -0,0 +1,130 @@
...
+    {
+    }
+
+    fn pop_namespace(&mut self) {}
+
+    fn get_root(&mut self) -> &mut Self::Root {
+        self
+    }
+
+    fn num_constraints(&self) -> usize {
```
kobigurk, line 76: Can you add similar methods to get the number of input and aux variables?

kobigurk (changes requested):

r1cs-std/src/test_constraint_counter.rs
gakonst: The fields are public, I just added this method to be consistent with `TestConstraintSystem`

r1cs-std/src/test_constraint_counter.rs
```diff
@@ -0,0 +1,130 @@
+use crate::String;
+use algebra::Field;
+use r1cs_core::{ConstraintSystem, Index, LinearCombination, SynthesisError, Variable};
+
+/// Constraint counter for testing purposes.
+pub struct TestConstraintCounter {
```
Pratyush:
```suggestion
pub struct ConstraintCounter {
```

r1cs-std/src/test_constraint_counter.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub num_inputs: usize,
+    pub num_aux: usize,
+    pub num_constraints: usize,
+}
+
+impl TestConstraintCounter {
+    pub fn new() -> TestConstraintCounter {
+        TestConstraintCounter {
```
Pratyush:
```suggestion
impl ConstraintCounter {
    pub fn new() -> Self {
        Self {
```

r1cs-std/src/test_constraint_counter.rs
```diff
@@ -0,0 +1,130 @@
...
+            num_constraints: 0,
+        }
+    }
+
+    pub fn num_constraints(&self) -> usize {
+        self.num_constraints
+    }
+}
+
+impl<ConstraintF: Field> ConstraintSystem<ConstraintF> for TestConstraintCounter {
```
Pratyush:
```suggestion
impl<ConstraintF: Field> ConstraintSystem<ConstraintF> for ConstraintCounter {
```

r1cs-std/src/test_constraint_counter.rs
```diff
@@ -0,0 +1,130 @@
...
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn test_constraints_counter() {
+        let empty_circuit = TestCircuit::<Fq>(None);
+        let populated_circuit = TestCircuit(Some(Fq::from(10u32)));
+
+        let mut counter = TestConstraintCounter::new();
```
Pratyush:
```suggestion
        let mut counter = ConstraintCounter::new();
```

gakonst:
FWIW CI is failing at nightly rustfmt which is not available

PR #153, opened 2020-03-26
use AllocGadget for UInt64

weikengchen:
UInt8 has implemented AllocGadget, but neither UInt32 nor UInt64 does. 

This PR implements AllocGadget for UInt64. The vanilla version of `alloc` is renamed as `_alloc`. 

In the future, ideally, UInt64 and UInt32 will share some methods with UInt8. Very likely, we need a trait named Int.

### DIFF
--- a/r1cs-std/src/bits/uint64.rs
+++ b/r1cs-std/src/bits/uint64.rs
@@ -338,13 +338,6 @@ impl<ConstraintF: Field> ToBytesGadget<ConstraintF> for UInt64 {
 
         Ok(bytes)
     }
-
-    fn to_bytes_strict<CS: ConstraintSystem<ConstraintF>>(
-        &self,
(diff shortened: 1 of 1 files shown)


Pratyush:
Yeah, I had a fork somewhere which unified all of these, but I haven't yet had a chance to update it

Pratyush:
I'll merge this once CI passes!

weikengchen:
It seems there is some upstream update that my local version does not have. Will boost my local to the new zexe and modify this PR.

weikengchen:
It quasi-passes the test. The remaining one seems to be a temporary error on Travis-CI.

```
component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel nightly
Sometimes not all components are available in any given nightly."
```

PR #159, opened 2020-03-30
Parallelize groth16

### LINKED ISSUE #157: Parallelize Groth16
Calculating a b and c can be done in parallel here:
https://github.com/scipr-lab/zexe/blob/master/groth16/src/prover.rs#L268-L332

Should improve our proving times. Implementation via crossbeam scoped threads should be straightforward.

I'll tackle this this week

gakonst:
fixes: https://github.com/scipr-lab/zexe/issues/157

As title, using Crossbeam, we spawn a thread and get its value via its `JoinHandle`.  I also added a new Error variant for Crossbeam Errors (that's where the `Box<dyn Any + Send>` comes in, not sure if there's a cleaner way to do this, have tried doing this before and couldn't find one)

<s>After changing the Groth16 example to use deterministic RNG, with 500k constraints and 25k inputs I get  ~15% improvements on the prover and the setup generator.</s>

EDIT: After running it multiple times, it seems that prover times actually are around 40sec for both versions. It _might_ be the case that all my CPU threads are occupied in Rayon's operations, not giving any room for this change to provide any benefit.

EDIT2: I've run the same benchmark with a larger circuit (5mil constraints, BLS12-377), and I get ~same performance as master. It probably is the case that FFTs are dominating the computation.

### DIFF
--- a/groth16/Cargo.toml
+++ b/groth16/Cargo.toml
@@ -35,6 +35,7 @@ crossbeam = "0.7.3"
 [dev-dependencies]
 csv = { version = "1" }
 algebra = { path = "../algebra", default-features = false, features = [ "bls12_377", "bls12_381", "sw6" ] }
+rand_xorshift = "0.2.0"
 
 [features]
 default = ["parallel"]
--- a/groth16/Cargo.toml
+++ b/groth16/Cargo.toml
@@ -30,6 +30,7 @@ r1cs-core = { path = "../r1cs-core", default-features = false }
 rand = { version = "0.7", default-features = false }
 rayon = { version = "1", optional = true }
 smallvec = "1.2.0"
+crossbeam = "0.7.3"
 
 [dev-dependencies]
 csv = { version = "1" }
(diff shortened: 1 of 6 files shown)


Pratyush:
Hmm this might run into issues with `no_std`.

gakonst:
I could abstract this and do conditional compilation to use crossbeam only if the `parallel` feature is turned on - let me try.

Pratyush:
Yeah, that sounds great!

kobigurk:
If you want to use this chance for more parallelization, it's worth parallelizing the miller loops in the verifier. Huge difference on mobile phones for SW6. I think `par_iter` should be enough for this.

Pratyush:
Hm to address the speedup issue, it might make sense to use `rayon::scope`; that would enable use of work-stealing? It might not help with the exhaustion of cores, but might help otherwise with better utilization?

gakonst:
OK on further thought: we might be CPU bound. 

I noticed that both when running the benchmark on master and on this branch my CPUs are at 100%. So perhaps further parallelism of this sort doesn't do much. We probably could focus more on the FFTs, @kobigurk says they shouldn't be CPU bound

gakonst:
Closing this due to the explanation given in: https://github.com/scipr-lab/zexe/pull/159#issuecomment-606064353

PR #163, opened 2020-04-02
Goff for mul_assign + field implementation macro + cheaper into_repr

jon-chuang:
Added goff (for `mul_assign` only) + macros to simplify code
Have not added requirements on the prime fields/alternate montgomery multiplication for primes not satisfying requirements (https://hackmd.io/@zkteam/modular_multiplication)

PR #167, opened 2020-04-03
Fall back to generic mul implementation if needed

jon-chuang:
Minor changes that should have been added to previous PR #163. 

By using my favourite tool of time `#[unroll_for_loops]`, I am able to get compile time checks that traverse the entire modulus, resulting in zero-overhead implementation switching. 

Alternate implementation to no carry is about 6% slower.

### DIFF
--- a/algebra-core/src/fields/macros.rs
+++ b/algebra-core/src/fields/macros.rs
@@ -432,27 +432,25 @@ macro_rules! impl_Fp {
 }
 
 
-// The no carry optimisation requires last bit to be zero
-// and all remaining bits not set
-// https://hackmd.io/@zkteam/modular_multiplication
-// It is slightly faster
+/// This modular multiplication algorithm uses Montgomery
+/// reduction for efficient implementation. It also additionally
+/// uses the "no-carry optimization" outlined
+/// [here](https://hackmd.io/@zkteam/modular_multiplication) if
+/// `P::MODULUS` has (a) a non-zero MSB, and (b) at least one
+/// zero bit in the rest of the modulus.
 macro_rules! impl_field_mul_assign {
     ($limbs:expr) => {
-        #[inline]
+        #[inline(never)]
         #[unroll_for_loops]
         fn mul_assign(&mut self, other: &Self) {
-            let mut no_carry = true;
-            if P::MODULUS.0[$limbs-1] >> 63 != 0 {
-                no_carry = false;
-            } else if ((P::MODULUS.0[$limbs-1] << 1) + 1) & !(0u64) == 0 {
-                let mut all_bits_set = true;
-                for i in 1..$limbs {
-                    if P::MODULUS.0[$limbs-1-i] & !(0u64) == 0 {
-                        all_bits_set &= true;
-                    } else { all_bits_set = false; }
-                }
-                no_carry = !all_bits_set;
+            // Checking the modulus at compile time
+            let first_bit_set = P::MODULUS.0[$limbs - 1] >> 63 != 0;
+            let mut all_bits_set = P::MODULUS.0[$limbs - 1] == !0 - (1 << 63);
+            for i in 1..$limbs {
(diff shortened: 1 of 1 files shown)


algebra-core/src/fields/macros.rs
```diff
@@ -431,27 +431,72 @@ macro_rules! impl_Fp {
...
+                no_carry = false;
+            } else if ((P::MODULUS.0[$limbs-1] << 1) + 1) & !(0u64) == 0 {
+                let mut all_bits_set = true;
+                for i in 1..$limbs {
+                    if P::MODULUS.0[$limbs-1-i] & !(0u64) == 0 {
+                        all_bits_set &= true;
+                    } else { all_bits_set = false; }
+                }
+                no_carry = !all_bits_set;
+            }
```
Pratyush: Just to ensure that it always gets compiled away, maybe we should wrap the logic in a `const fn`? In this case maybe we can move the `const fn` out of the function. I also included some simplifications of the logic, but you should double-check that these  simplifications are actually correct.

```suggestion
			#[inline(always)]
			#[unroll_for_loops]
			const fn can_use_no_carry_opt(modulus: [u64; $limbs]) -> bool {
	            if (P::MODULUS.0[$limbs - 1] >> 63) != 0 {
	                false
	            } else if (!P::MODULUS.0[$limbs - 1]) == (1 << 63) {
	            	// after flipping the bits of the last limb, every bit 
	            	// except the first was set
	            	
	                let mut all_bits_set = true;
	                for i in 1..$limbs {
	                    // the limb is equal to all-ones.
	                    all_bits_set &= (P::MODULUS.0[$limbs - i - 1] == !0u64);
	                }
	                !all_bits_set
	            } else {
	            	true
	            }
			}
			
			let no_carry = can_use_no_carry_opt(P::MODULUS.0);
```

algebra-core/src/fields/macros.rs
```diff
@@ -431,27 +431,72 @@ macro_rules! impl_Fp {
     }
 }
 
+
+// The no carry optimisation requires last bit to be zero
+// and all remaining bits not set
+// https://hackmd.io/@zkteam/modular_multiplication
+// It is slightly faster
```
Pratyush:
```suggestion
/// This modular multiplication algorithm uses Montgomery
/// reduction for efficient implementation. It also additionally
/// uses the "no-carry optimization" outlined
/// [here](https://hackmd.io/@zkteam/modular_multiplication) if
/// `P::MODULUS` has (a) a non-zero MSB, and (b) at least one
/// zero bit in the rest of the modulus. 
```

algebra-core/src/fields/macros.rs
```diff
@@ -431,27 +431,72 @@ macro_rules! impl_Fp {
...
+            } else if ((P::MODULUS.0[$limbs-1] << 1) + 1) & !(0u64) == 0 {
+                let mut all_bits_set = true;
+                for i in 1..$limbs {
+                    if P::MODULUS.0[$limbs-1-i] & !(0u64) == 0 {
+                        all_bits_set &= true;
+                    } else { all_bits_set = false; }
+                }
+                no_carry = !all_bits_set;
+            }
+            // No carry optimisation applied to CIOS
```
Pratyush:
```suggestion
            // No-carry optimization applied to CIOS
```

Pratyush (commented):
Thanks, left a few small comments.

algebra-core/src/fields/macros.rs
jon-chuang: Unfortunately, due to the nature of `unroll_for_loops`, loops and functions containing loops are not seen as constant by the compiler. 

I have opened an issue https://gitlab.com/elrnv/unroll/-/issues/2 but it won't be moving for a while at least, I suspect. 

For now, I suggest to believe in the compiler. Constant propogation is the most basic compiler optimisation :)

PR #171, opened 2020-04-04
Optimize Groth16 memory and run-time performance

gakonst:
Turns out we were making redundant evaluations? The evaluation of the polynomial was already being done in R1CS-to-QAP -- cc @kobigurk

### DIFF
--- a/groth16/Cargo.toml
+++ b/groth16/Cargo.toml
@@ -29,7 +29,6 @@ ff-fft = { path = "../ff-fft", default-features = false }
 r1cs-core = { path = "../r1cs-core", default-features = false }
 rand = { version = "0.7", default-features = false }
 rayon = { version = "1", optional = true }
-smallvec = "1.2.0"
 
 [dev-dependencies]
 csv = { version = "1" }
--- a/groth16/examples/snark-scalability/groth16.rs
+++ b/groth16/examples/snark-scalability/groth16.rs
@@ -128,7 +128,8 @@ fn main() -> Result<(), Box<dyn Error>> {
         let start = Instant::now();
         // let proof = Proof::read(&proof_vec[..]).unwrap();
         // Check the proof
-        let _ = verify_proof(&pvk, &proof, &inputs).unwrap();
+        let r = verify_proof(&pvk, &proof, &inputs).unwrap();
+        assert!(r);
         total_verifying += start.elapsed();
     }
 
--- a/groth16/examples/snark-scalability/groth16.rs
+++ b/groth16/examples/snark-scalability/groth16.rs
@@ -145,15 +145,14 @@ fn main() -> Result<(), Box<dyn Error>> {
         verifying_avg.subsec_nanos() as f64 / 1_000_000_000f64 + (verifying_avg.as_secs() as f64);
 
     println!(
-        "=== Benchmarking Groth16 with {} inputs and {} constraints: ====",
-        num_inputs, num_constraints
+        "=== Benchmarking Groth16 with {} constraints: ====",
+        num_constraints
     );
     println!("Average setup time: {:?} seconds", setup_avg);
     println!("Average proving time: {:?} seconds", proving_avg);
     println!("Average verifying time: {:?} seconds", verifying_avg);
 
     wtr.write_record(&[
-        format!("{}", num_inputs),
         format!("{}", num_constraints),
         format!("{}", setup_avg),
         format!("{}", proving_avg),
--- a/groth16/examples/snark-scalability/groth16.rs
+++ b/groth16/examples/snark-scalability/groth16.rs
@@ -52,14 +52,13 @@ use crate::constraints::Benchmark;
 
 fn main() -> Result<(), Box<dyn Error>> {
     let args: Vec<String> = env::args().collect();
-    if args.len() < 4 || args[1] == "-h" || args[1] == "--help" {
+    if args.len() < 3 || args[1] == "-h" || args[1] == "--help" {
         println!(
-            "\nHelp: Invoke this as <program> <num_inputs> <num_constraints> <output_file_path>\n"
+            "\nHelp: Invoke this as <program> <num_constraints> <output_file_path>\n"
         );
     }
-    let num_inputs: usize = args[1].parse().unwrap();
-    let num_constraints: usize = args[2].parse().unwrap();
-    let output_file_path = PathBuf::from(args[3].clone());
+    let num_constraints: usize = args[1].parse().unwrap();
+    let output_file_path = PathBuf::from(args[2].clone());
     let mut wtr = if !output_file_path.exists() {
         println!("Creating output file");
         let f = OpenOptions::new()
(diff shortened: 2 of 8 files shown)


groth16/src/generator.rs
```diff
@@ -86,34 +86,22 @@ impl<E: PairingEngine> ConstraintSystem<E::Fr> for KeypairAssembly<E> {
         LB: FnOnce(LinearCombination<E::Fr>) -> LinearCombination<E::Fr>,
         LC: FnOnce(LinearCombination<E::Fr>) -> LinearCombination<E::Fr>,
     {
-        fn eval<E: PairingEngine>(
```
gakonst, line 89: This was never really a polynomial evaluation so I moved and renamed it, so that we can share functionality between the prover/generator files

groth16/examples/snark-scalability/groth16.rs
```diff
@@ -52,14 +52,13 @@ use crate::constraints::Benchmark;
 
 fn main() -> Result<(), Box<dyn Error>> {
     let args: Vec<String> = env::args().collect();
-    if args.len() < 4 || args[1] == "-h" || args[1] == "--help" {
+    if args.len() < 3 || args[1] == "-h" || args[1] == "--help" {
         println!(
-            "\nHelp: Invoke this as <program> <num_inputs> <num_constraints> <output_file_path>\n"
+            "\nHelp: Invoke this as <program> <num_constraints> <output_file_path>\n"
```
gakonst, line 57: This was not used anywhere

groth16/Cargo.toml
```diff
@@ -29,7 +29,6 @@ ff-fft = { path = "../ff-fft", default-features = false }
 r1cs-core = { path = "../r1cs-core", default-features = false }
 rand = { version = "0.7", default-features = false }
 rayon = { version = "1", optional = true }
-smallvec = "1.2.0"
```
gakonst, line 32: Did this provide any real benefit? It was only used for SmallVecs of size 2, which we'd instantly exceed after 2 `enforce` calls

groth16/src/prover.rs
```diff
@@ -272,3 +266,25 @@ where
         c: g_c.into_affine(),
     })
 }
+
+fn calculate_coeff<G: AffineCurve>(
```
gakonst, line 232: abstracted coefficient calculation to a function - reduces error chances from copy-pasting

groth16/src/prover.rs
```diff
@@ -242,58 +155,63 @@ where
     let (full_input_assignment, h, _) = R1CStoQAP::witness_map::<E>(&prover)?;
     end_timer!(witness_map_time);
 
-    let input_assignment = full_input_assignment[1..prover.num_inputs]
-        .into_iter()
+    let num_inputs = prover.input_assignment.len();
+
+    let input_assignment = cfg_into_iter!(full_input_assignment[1..num_inputs])
```
Pratyush: Because inputs are usually small, and the into_repr computation pretty cheap, this wouldn’t really help, and might even harm things because of the overhead due to synchronization.

groth16/src/r1cs_to_qap.rs
```diff
@@ -129,7 +128,7 @@ impl R1CStoQAP {
             });
 
         for i in 0..num_inputs {
-            a[num_constraints + i] = if i > 0 { full_input_assignment[i] } else { one };
+            a[num_constraints + i] = full_input_assignment[i];
```
gakonst, line 138: I believe this was not needed?

groth16/src/r1cs_to_qap.rs
```diff
@@ -159,11 +158,6 @@ impl R1CStoQAP {
         domain.divide_by_vanishing_poly_on_coset_in_place(&mut ab);
         domain.coset_ifft_in_place(&mut ab);
 
-        let mut h: Vec<E::Fr> = vec![zero; domain_size - 1];
```
gakonst, line 150: This was redundant as well, I think?

groth16/src/prover.rs
gakonst: Good point, fixed https://github.com/scipr-lab/zexe/pull/171/commits/2d9fe6ab08593cc44f578686b6787d9b5acd9343#diff-4a7e8a6b598180387474d3d2ad21ab90R162-R163

groth16/src/r1cs_to_qap.rs
```diff
@@ -1,13 +1,45 @@
...
+#[inline]
+fn evaluate_constraint<'a, LHS, RHS, R>(
+    terms: &'a [(LHS, Index)],
+    assignment: &'a [RHS],
+    num_inputs: usize,
+) -> R
+where
+    LHS: One + Send + Sync + PartialEq,
+    RHS: Send + Sync + std::ops::Mul<&'a LHS, Output = RHS> + Copy,
+    R: std::iter::Sum + Zero + Send + AddAssign<RHS>,
```
Pratyush, line 20:
```suggestion
    R: core::iter::Sum + Zero + Send + AddAssign<RHS>,
```

Pratyush:
```suggestion
    RHS: Send + Sync + core::ops::Mul<&'a LHS, Output = RHS> + Copy,
    R: core::iter::Sum + Zero + Send + AddAssign<RHS>,
```

Pratyush:
Your changes should reduce max memory consumption, right? nice!

gakonst:
@Pratyush yes. They also remove some arithmetic which was being done redundantly 
 [here](https://github.com/scipr-lab/zexe/pull/171/files#diff-4a7e8a6b598180387474d3d2ad21ab90L46-L54)

groth16/src/r1cs_to_qap.rs
```diff
@@ -8,6 +8,38 @@ use r1cs_core::{ConstraintSystem, Index, SynthesisError};
 #[cfg(feature = "parallel")]
 use rayon::prelude::*;
 
+#[inline]
```
gakonst, line 11: This function could be worth moving to `r1cs-core` or `r1cs-std`? It's pretty abstract now, and we could reuse it, in GM17's `witness_map` as well

e.g. we could remove our function from the [phase2-mpc](https://github.com/celo-org/snark-setup/blob/master/phase2/src/polynomial.rs#L94-L116)

gakonst:
Also, on a closer look: it seems to me that the same redundancies exist in GM17. Unsure who uses it at the moment, but we should also port them there at some point

groth16/src/prover.rs
gakonst: Is there maybe a better name for this function? Trying to think where also we might see it, beyond groth16 or gm17

groth16/src/r1cs_to_qap.rs
Pratyush: Yeah I also have a similar function in the Marlin codebase. I intend on doing a refactor of the CS infrastructure starting within a couple of weeks, so I’ll do this as part of that!

Pratyush:
@gakonst is this ready for merging?

gakonst:
Yes, @Pratyush, go ahead!

PR #177, opened 2020-04-08
Fix COFACTORs of new MNT curves

paberr:
While experimenting with my mixed-radix FFT implementation, I found that I incorrectly computed the G2 `COFACTOR` constant for the new MNT4/6 curves.
I accidentally multiplied them by `R` mod `MODULUS`.

I fixed my [constant generation script](https://github.com/paberr/mnt4-mnt6-scripts/commit/5d1ef9850802196a529e819b45b9b4631acb2aad) and updated the `COFACTOR`s accordingly.

Also, I added a few use statements for `Vec` that prevented compilation with `no_std`.


Pratyush:
Thanks for the PR!

It's surprising that CI didn't catch this; I made a PR #178 to test more code in `no_std`, but shouldn't the tests have caught the cofactor issue?

paberr:
> Thanks for the PR!
> 
> It's surprising that CI didn't catch this; I made a PR #178 to test more code in `no_std`, but shouldn't the tests have caught the cofactor issue?

Interesting question, yes. Ideally, they should but I think the `mnt6_753_g2_gadget_test` doesn't cover that.
I only discovered it, because `G2Gadget::alloc` and `G2Gadget::alloc_checked` returned different results (the latter uses the `COFACTOR` constant).
I can add a test that will cover this issue in the future.

Pratyush:
It's still surprising that even the non-gadget tests didn't catch this. I'd appreciate it if you could add tests for this functionality in the native code also =)

paberr:
True. I added tests for the COFACTOR in the native case, generalised the tests for the gadgets, and added a test for the `alloc_checked`.

As expected, the tests fail prior to my fixes and pass afterwards.

Failures before my fixes:
```
    mnt4_298::curves::tests::test_g2_projective_curve
    mnt4_753::curves::tests::test_g2_projective_curve
    mnt6_753::curves::tests::test_g2_projective_curve
    instantiated::mnt6_753::curves::test
```
*Note:* mnt6_298 had the correct COFACTOR assigned.
Only the mnt6_753 curve fails with the `alloc_checked` test, because the the 298 bit curves do not use the COFACTOR in `alloc_checked`.

**EDIT:** I also rebased to the current master to get rid of the cargo fmt issues in the CI.

Pratyush:
> because the the 298 bit curves do not use the COFACTOR in alloc_checked

This is because they just multiply by the scalar field modulus, right?

paberr:
> This is because they just multiply by the scalar field modulus, right?

Exactly.

PR #179, opened 2020-04-09
Mixed-radix FFT & Recursive SNARKs example

paberr:
This PR builds & depends on #177.

### Motivation for the mixed-radix FFT
The radix-2 FFT that currently comes with zexe only works for a number of constraints + inputs below 2^TWO_ADICITY of the Fr field one is proving on (considering Groth16). For MNT6-753 that means we can have a maximum of 2^15 (constraints+inputs). Unfortunately, this is not enough to verify another Groth16 proof in a circuit.

The mixed-radix FFT allows to increase this number to 2^15*5^2 for MNT6-753 using a small subgroup with base 5 and power 2, which is more than enough for the previous motivational example.

### Other implementations
Inspired by https://github.com/o1-labs/snarky/pull/107, which adds a mixed-radix FFT to libfqfft, I added a mixed-radix FFT to zexe's FFT implementation.
It turned out that ginger-lib implemented the same mixed-radix FFT concurrently here: https://github.com/ZencashOfficial/ginger-lib/pull/34

The main difference to the ginger-lib implementation is that they split up the two FFTs into separate structs, while I followed Snarky's example and put both of them into the single `EvaluationDomain` struct we have right now.

### Preliminary results
I already ran some examples using recursive SNARKs on the MNT4/6-753 curves and here are some results:
```
MNT6_753 curve, 151,138 constraints, mixed-radix FFT
Average setup time: 80.412886252 seconds
Average proving time: 54.170403498 seconds

MNT4_753 curve, 276,480 constraints, radix-2 FFT
Average setup time: 104.938389255 seconds
Average proving time: 83.827640713 seconds
```
While the numbers are not directly comparable, they suggest that the mixed-radix FFT (at least for this subgroup) doesn't add a substantial overhead.
However, I also want to implement more comparable benchmarks (see TODOs).

### TODOs
Since we might want to add other FFT variants in the future (e.g., using the radix-2 with multiple cosets), I still want to do the following improvements before this is ready to merge:

- [x] Split up the two FFT variants while avoiding code duplication (so something similar to what ginger-lib did).
- [x] Add benchmarks for the FFT variants

More feedback welcome. :)


ff-fft/src/domain/mod.rs
```diff
@@ -81,20 +88,62 @@ impl<F: PrimeField> EvaluationDomain<F> {
...
+
+            group_gen = F::get_root_of_unity(num_coeffs)?;
+        } else {
+            // Compute the size of our evaluation domain
+            size = num_coeffs.next_power_of_two() as u64;
+            log_size_of_group = size.trailing_zeros();
+
+            // TODO: Check > vs. >= here.
+            // libfqfft uses > https://github.com/scipr-lab/libfqfft/blob/e0183b2cef7d4c5deb21a6eaf3fe3b586d738fe0/libfqfft/evaluation_domain/domains/basic_radix2_domain.tcc#L33
+            if log_size_of_group > F::Params::TWO_ADICITY {
```
paberr: This should be re-checked. Zexe's implementation previously did not allow `log_size_of_group` to be equal to the 2-adicity. libfqfft does allow that, though.

algebra-core/src/fields/mod.rs
```diff
@@ -193,6 +195,14 @@ pub trait FpParameters: 'static + Send + Sync + Sized {
...
     /// (Self::MODULUS - 1) / 2
     const MODULUS_MINUS_ONE_DIV_TWO: Self::BigInt;
+
+    /// The base of a small subgroup that can be used for mixed-radix FFT.
+    const SMALL_SUBGROUP_BASE: Option<u32> = None;
+    /// The power of a small subgroup that can be used for mixed-radix FFT.
+    const SMALL_SUBGROUP_POWER: Option<u32> = None;
+    /// GENERATOR^((MODULUS-1) / (2^s *
+    /// SMALL_SUBGROUP_BASE^SMALL_SUBGROUP_POWER)) Used for mixed-radix FFT.
+    const FULL_ROOT_OF_UNITY: Option<Self::BigInt> = None;
```
Pratyush: As I mentioned on Telegram, maybe we should move these parameters, along with the `ROOT_OF_UNITY`-related stuff to a separate `FftField: Field` trait?

Then we require `PrimeField: FftField`?

This way we can change the EvaluationDomain to only care about the FFT structure, so that it becomes `EvaluationDomain<F: FftField>`.

groth16/Cargo.toml
```diff
@@ -32,7 +32,9 @@ rayon = { version = "1", optional = true }
 
 [dev-dependencies]
 csv = { version = "1" }
-algebra = { path = "../algebra", default-features = false, features = [ "bls12_377", "bls12_381", "sw6" ] }
+algebra = { path = "../algebra", default-features = false, features = [ "bls12_377", "bls12_381", "sw6", "mnt6_753", "mnt4_753" ] }
+r1cs-std = { path = "../r1cs-std", default-features = false, features = [ "mnt6_753", "mnt4_753" ] }
+crypto-primitives = { path = "../crypto-primitives", default-features = false, features = [ "std", "groth16", "r1cs" ] }
```
Pratyush:
```suggestion
crypto-primitives = { path = "../crypto-primitives", default-features = false, features = [ "groth16", "r1cs" ] }
```

gm17/Cargo.toml
```diff
@@ -32,7 +32,9 @@ smallvec = "1.2"
 
 [dev-dependencies]
 csv = { version = "1" }
-algebra = { path = "../algebra", default-features = false, features = [ "bls12_377", "bls12_381", "sw6" ] }
+algebra = { path = "../algebra", default-features = false, features = [ "bls12_377", "bls12_381", "sw6", "mnt6_753", "mnt4_753" ] }
+r1cs-std = { path = "../r1cs-std", default-features = false, features = [ "mnt6_753", "mnt4_753" ] }
+crypto-primitives = { path = "../crypto-primitives", default-features = false, features = [ "std", "gm17", "r1cs" ] }
```
Pratyush:
```suggestion
crypto-primitives = { path = "../crypto-primitives", default-features = false, features = [ "gm17", "r1cs" ] }
```

ff-fft/src/lib.rs
```diff
@@ -15,9 +15,11 @@ extern crate alloc;
 
 #[cfg(not(feature = "std"))]
 pub(crate) use alloc::{borrow::Cow, collections::BTreeMap, vec::Vec};
+#[cfg(not(feature = "std"))]
+pub(crate) use core::cmp::min;
 
 #[cfg(feature = "std")]
-pub(crate) use std::{borrow::Cow, collections::BTreeMap, vec::Vec};
+pub(crate) use std::{borrow::Cow, cmp::min, collections::BTreeMap, vec::Vec};
```
Pratyush: You can import `core::cmp::min` even when `no_std` is off, so you don't need to make this change =)

algebra-core/src/fields/mod.rs
paberr: Would you put `TWO_ADICITY` also into the `FftField` trait?
As far as I can see, it is mostly used in the FFT related stuff with the exception of one occurrence in the `sqrt_impl`. But then, the `sqrt_impl` also implicitly relies on the `ROOT_OF_UNITY` (which is the default for `qnr_to_t`).

paberr: Also, I think there are three options here:
1. Putting the methods and constants into the `FftField` trait, which requires to implement this trait for our prime fields (just as we do with FpParameters right now).
2. Putting only the methods into the `FftField` trait, so that the `FftField` trait can be implemented in the `impl_Fp` macro.
3. Actually having a second trait like `FftParameters`, which has the constants and needs to be implemented alongside `FpParameters`. The `FftField` trait then works analogously to the `PrimeField` trait in that it has a `Params` type. The methods of the `FftField` trait can then be implemented in the `impl_Fp` macro.

The disadvantage of the first option is that methods like `full_root_of_unity` cannot have a default implementation as there is no way of creating a `Self` from just a `Self::BigInt` at this point.

Pratyush: I think option 3 is the way to go, though it is a little bit more work than the other options. We can make `FpParameters: FftParameters`, and `PrimeField: FftField`, with the appropriate bound on the Parameters associated type

Pratyush: And yeah, I’d put TWO_ADICITY and related constants in the new parameters trait

paberr: It turns out that splitting these traits is not easily achievable without any downsides.

The problem arises from the following constraints:
1. For the ff-fft crate to be able to work, it requires `F: FftField` and `FftField` must have a type parameter `FftParams: FftParameters`.
2. We would like that `PrimeField: FftField`. Since, `PrimeField` requires a type parameter `Params: FpParameters`, we end up with something similar to the following:
```rust
trait FftParameters {
    type BigInt: BigInteger;
    ...
}
trait FpParameters: FftParameters { ... }
trait FftField: Field {
    type FftParams: FftParameters;
}
trait PrimeField: FftField {
     type Params: FpParameters<BigInt = Self::BigInt>;
     type BigInt: BigInteger;
}
```
This still works fine. However, we would like to require the two BigInts to be the same or even `FftParams = Params` for the `PrimeField`.
Doing something like `trait PrimeField: FftField<FftParams = Self::Params>`, results in a cyclic dependency though ([error[E0391]](https://doc.rust-lang.org/error-index.html#E0391)).

So far, I haven't found a satisfactory solution splitting these traits. So, I'm not sure this is worth it unless you have another idea that I might be missing here. :)

Pratyush: I think if you do `PrimeField: FftField<FftParams = <Self as PrimeField>::Params>`, then it should work?

paberr: You're completely right. Perfect, then this will work out. :)

ff-fft/src/domain/mod.rs
paberr: As far as I can see, there is no reason preventing the FFT code to run with the 2^s root of unity, which is a normal root of unity after all. The code for `get_root_of_unity` also handles this case fine.

paberr:
I addressed all TODOs and rebased to the current master.

paberr:
Btw. it seems that Travis times out when building the branch.
Any idea how to fix that? @Pratyush

paberr:
Travis is now passing. I have no idea why it doesn't show up in the PR, but here it is: https://travis-ci.com/github/nimiq/zexe/builds/161205189

algebra-core/src/fields/mod.rs
```diff
@@ -158,10 +160,30 @@ pub trait Field:
...
+/// A trait that defines parameters for a field that can be used for FFT.
+pub trait FftParameters: 'static + Send + Sync + Sized {
     type BigInt: BigInteger;
 
+    /// 2^s * t = MODULUS - 1 with t odd. This is the two-adicity of
+    /// `Self::MODULUS`.
+    const TWO_ADICITY: u32;
+
+    /// 2^s root of unity computed by GENERATOR^t
+    const ROOT_OF_UNITY: Self::BigInt;
```
Pratyush, line 172: Can this just be `Self`?
```suggestion
    /// 2^s root of unity computed by GENERATOR^t
    const ROOT_OF_UNITY: Self;
```

algebra-core/src/fields/mod.rs
```diff
@@ -158,10 +160,30 @@ pub trait Field:
...
+
+    /// The base of a small subgroup that can be used for mixed-radix FFT.
+    const SMALL_SUBGROUP_BASE: Option<u32> = None;
+
+    /// The power of a small subgroup that can be used for mixed-radix FFT.
+    const SMALL_SUBGROUP_POWER: Option<u32> = None;
+
+    /// GENERATOR^((MODULUS-1) / (2^s *
+    /// SMALL_SUBGROUP_BASE^SMALL_SUBGROUP_POWER)) Used for mixed-radix FFT.
+    const FULL_ROOT_OF_UNITY: Option<Self::BigInt> = None;
```
Pratyush, line 182: Same here?

```suggestion
    /// SMALL_SUBGROUP_BASE^SMALL_SUBGROUP_POWER)) Used for mixed-radix FFT.
    const FULL_ROOT_OF_UNITY: Option<Self> = None;
```

ff-fft-benches/LICENSE-APACHE
```diff
@@ -0,0 +1,201 @@
+                              Apache License
```
Pratyush: This should be a symlink, like in the other crates

ff-fft-benches/LICENSE-MIT
```diff
@@ -0,0 +1,19 @@
+The MIT License (MIT)
```
Pratyush: Same here

ff-fft/src/domain/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+//! the scalar field.
+//!
+//! In pairing-based SNARKs like GM17, we need to calculate
+//! a quotient polynomial over a target polynomial with roots
+//! at distinct points associated with each constraint of the
+//! constraint system. In order to be efficient, we try to choose these
+//! roots to be the powers of a 2^n root of unity in the field.
+//! This allows us to perform polynomial operations in O(n)
+//! by performing an O(n log n) FFT over such a domain.
+//!
```
Pratyush:
```suggestion
//! This module contains an `EvaluationDomain` abstraction for
//! performing various kinds of polynomial arithmetic on top of
//! fields that are friendly to fast-fourier-transforms (FFTs).
//!
//! A field is FFT-friendly if it contains enough 
//! roots of unity to perform the FFT in O(n log n) time.
//! These roots of unity comprise the domain over which
//! polynomial arithmetic is performed.
//!
```

ff-fft/src/domain/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+//! If the 2-adicity of the field is too small, but a small subgroup
+//! over the field is defined, we can try to build a mixed-radix evaluation
```
Pratyush:
```suggestion
//! over the field is defined, we try to build a mixed-radix evaluation
```

ff-fft/src/domain/radix2.rs
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    fn compute_size_of_domain(num_coeffs: usize) -> Option<usize> {
+        let size = num_coeffs.next_power_of_two();
+        // TODO: Check > vs. >= here.
+        if size.trailing_zeros() <= F::FftParams::TWO_ADICITY {
+            Some(size)
+        } else {
+            None
+        }
```
Pratyush: I think it's correct to have `<=`, but let's make it consistent with the equivalent line in `new`, so that we use `>` in both places:

```suggestion
        let size = num_coeffs.next_power_of_two();
        if size.trailing_zeros() > F::FftParams::TWO_ADICITY {
            None
        } else {
            Some(size)
        }
```

ff-fft/src/domain/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+pub mod radix2;
+pub(crate) mod utils;
+
+pub use general::GeneralEvaluationDomain;
+pub use mixed_radix::MixedRadixEvaluationDomain;
+pub use radix2::Radix2EvaluationDomain;
+
+/// Defines a domain over which finite field (I)FFTs can be performed. Works
+/// only for fields that have a large multiplicative subgroup of size that is
+/// a power-of-2.
```
Pratyush: This is no longer accurate, right?
```suggestion
/// Defines a domain over which finite field (I)FFTs can be performed. The
/// size of the supported FFT depends on the size of the multiplicative 
/// subgroup. For efficiency, we recommend that the field has at least one large
/// subgroup generated by a root of unity.
```

ff-fft/src/domain/mixed_radix.rs
```diff
@@ -0,0 +1,475 @@
+//! This module implements the `EvaluationDomain` abstraction for
+//! performing various kinds of polynomial arithmetic on top of
+//! the scalar field using a mixed-radix domain.
```
Pratyush:
```suggestion
//! This module contains a `MixedRadixEvaluationDomain` for
//! performing various kinds of polynomial arithmetic on top of
//! fields that are FFT-friendly but do not have high-enough 
//! two-adicity to perform the FFT efficiently, i.e. the multiplicative
//! subgroup `G` generated by `F::ROOT_OF_UNITY` is not large enough.
//! `MixedRadixEvaluationDomain` resolves
//! this issue by using a larger subgroup obtained by combining
//! `G` with another subgroup of size 
//! `F::SMALL_SUBGROUP_BASE^(F::SMALL_SUBGROUP_POWER)`, 
//! to obtain a subgroup generated by `F::FULL_ROOT_OF_UNITY`.
```

(We should change the constant names here if we update them in the `FftParameters` trait.)

ff-fft/src/domain/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+    fn ifft<T: DomainCoeff<F>>(&self, evals: &[T]) -> Vec<T> {
+        let mut evals = evals.to_vec();
+        self.ifft_in_place(&mut evals);
+        evals
+    }
+
+    /// Compute a IFFT, modifying the vector in place.
+    fn ifft_in_place<T: DomainCoeff<F>>(&self, evals: &mut Vec<T>);
+
+    /// Distribute the powers of `g` among the `coeffs`.
```
Pratyush:
```suggestion
    /// Multiply the `i`-th element of `coeffs` with the `i`-th power of `g`.
```

algebra-core/src/fields/mod.rs
```diff
@@ -158,10 +160,30 @@ pub trait Field:
...
+    const TWO_ADICITY: u32;
+
+    /// 2^s root of unity computed by GENERATOR^t
+    const ROOT_OF_UNITY: Self::BigInt;
+
+    /// The base of a small subgroup that can be used for mixed-radix FFT.
+    const SMALL_SUBGROUP_BASE: Option<u32> = None;
+
+    /// The power of a small subgroup that can be used for mixed-radix FFT.
+    const SMALL_SUBGROUP_POWER: Option<u32> = None;
```
Pratyush:
```suggestion
    /// The integer `k` such that there exists a multiplicative subgroup
    /// of size `Self::SMALL_SUBGROUP_BASE^k`.
    const SMALL_SUBGROUP_POWER: Option<u32> = None;
```
Maybe we should also rename this to be analagous to `TWO_ADICITY`?

algebra-core/src/fields/mod.rs
```diff
@@ -158,10 +160,30 @@ pub trait Field:
...
 
+    /// 2^s * t = MODULUS - 1 with t odd. This is the two-adicity of
+    /// `Self::MODULUS`.
+    const TWO_ADICITY: u32;
+
+    /// 2^s root of unity computed by GENERATOR^t
+    const ROOT_OF_UNITY: Self::BigInt;
+
+    /// The base of a small subgroup that can be used for mixed-radix FFT.
+    const SMALL_SUBGROUP_BASE: Option<u32> = None;
```
Pratyush:
```suggestion
    /// An integer `b` such that there exists a multiplicative subgroup
    /// of size `b^k` for some integer `k`.
    const SMALL_SUBGROUP_BASE: Option<u32> = None;
```

algebra-core/src/fields/mod.rs
```diff
@@ -158,10 +160,30 @@ pub trait Field:
...
 
-/// A trait that defines parameters for a prime field.
-pub trait FpParameters: 'static + Send + Sync + Sized {
+/// A trait that defines parameters for a field that can be used for FFT.
+pub trait FftParameters: 'static + Send + Sync + Sized {
     type BigInt: BigInteger;
 
+    /// 2^s * t = MODULUS - 1 with t odd. This is the two-adicity of
+    /// `Self::MODULUS`.
+    const TWO_ADICITY: u32;
```
Pratyush:
```suggestion
	/// Let `N` be the size of the multiplicative group defined by the field.
	/// Then `TWO_ADICITY` is the two-adicity of `N`, i.e. the integer `s` 
    /// such that `N = 2^s * t` for some odd integer `t`.
    const TWO_ADICITY: u32;
```

algebra-core/src/fields/mod.rs
```diff
@@ -191,50 +213,99 @@ pub trait FpParameters: 'static + Send + Sync + Sized {
...
-    const TWO_ADICITY: u32;
-
     /// t for 2^s * t = MODULUS - 1
     const T: Self::BigInt;
 
-    /// 2^s root of unity computed by GENERATOR^t
-    const ROOT_OF_UNITY: Self::BigInt;
-
     /// (t - 1) / 2
     const T_MINUS_ONE_DIV_TWO: Self::BigInt;
```
Pratyush, line 223: Should we move `T` and `T_MINUS_ONE_DIV_TWO` also to `FftParameters`?

algebra-core/src/fields/mod.rs
```diff
@@ -158,10 +160,30 @@ pub trait Field:
     }
 }
 
-/// A trait that defines parameters for a prime field.
-pub trait FpParameters: 'static + Send + Sync + Sized {
+/// A trait that defines parameters for a field that can be used for FFT.
```
Pratyush:
```suggestion
/// A trait that defines parameters for a field that can be used for FFTs.
```

algebra-core/src/fields/mod.rs
```diff
@@ -191,50 +213,99 @@ pub trait FpParameters: 'static + Send + Sync + Sized {
...
+
+    /// Returns the 2^s * small_subgroup_base^small_subgroup_power root of unity
+    /// if a small subgroup is defined.
+    fn full_root_of_unity() -> Option<Self>;
+
+    /// Returns the multiplicative generator of `char()` - 1 order.
+    fn multiplicative_generator() -> Self;
+
+    /// Returns the root of unity of order n (for n a power of 2), if one
+    /// exists.
```
Pratyush: This comment isn't accurate any more, right?

algebra-core/src/fields/mod.rs
Pratyush: Maybe for clarity we should rename this to `TWO_ADIC_ROOT_OF_UNITY` or `TWO_ADIC_SUBGROUP_GENERATOR`? Currently it's not clear from the name what the difference is between `ROOT_OF_UNITY` and `FULL_ROOT_OF_UNITY`

algebra-core/src/fields/mod.rs
Pratyush: Like above, maybe we should change this to `LARGE_SUBGROUP_ROOT_OF_UNITY`/`LARGE_SUBGROUP_GENERATOR`?

algebra-core/src/fields/mod.rs
```diff
@@ -191,50 +213,99 @@ pub trait FpParameters: 'static + Send + Sync + Sized {
...
+/// The interface for fields that are able to be used in FFTs.
+pub trait FftField: Field + From<u128> + From<u64> + From<u32> + From<u16> + From<u8> {
+    type FftParams: FftParameters;
+
+    /// Returns the 2^s root of unity.
+    fn root_of_unity() -> Self;
+
+    /// Returns the 2^s * small_subgroup_base^small_subgroup_power root of unity
+    /// if a small subgroup is defined.
+    fn full_root_of_unity() -> Option<Self>;
```
Pratyush: If we rename the constants, we should rename these  methods as well.

ff-fft/src/domain/radix2.rs
```diff
@@ -0,0 +1,312 @@
+//! This module implements the `EvaluationDomain` abstraction for
+//! performing various kinds of polynomial arithmetic on top of
+//! the scalar field using a radix-2 domain.
```
Pratyush:
```suggestion
//! This module defines `Radix2EvaluationDomain`, an `EvaluationDomain`
//! for performing various kinds of polynomial arithmetic on top of
//! fields that are FFT-friendly. `Radix2EvaluationDomain` supports
//! FFTs of size at most `2^F::TWO_ADICITY`.
```

ff-fft/src/evaluations.rs
```diff
@@ -1,21 +1,21 @@
...
 
 use crate::{DensePolynomial, EvaluationDomain, Vec};
-use algebra_core::PrimeField;
+use algebra_core::FftField;
 use core::ops::{Add, AddAssign, Div, DivAssign, Index, Mul, MulAssign, Sub, SubAssign};
 
 /// Stores a polynomial in evaluation form.
 #[derive(Clone, PartialEq, Eq, Hash, Debug)]
-pub struct Evaluations<F: PrimeField> {
+pub struct Evaluations<F: FftField, D: EvaluationDomain<F>> {
```
Pratyush: Should we set the default type `D = GeneralEvaluationDomain<F>`?

ff-fft/src/domain/radix2.rs
paberr: Good point, also I forgot to remove the TODO here. :)

ff-fft/src/domain/mod.rs
paberr: Completely true. Thanks for taking care of the comments. :)

algebra-core/src/fields/mod.rs
paberr: Yes, I like that idea! I'll go with the `_ROOT_OF_UNITY` versions to be a bit more consistent with other implementations. I guess that will make it easier for others to compare the code.

ff-fft/src/domain/mod.rs
paberr: That's indeed much more accurate.

algebra-core/src/fields/mod.rs
paberr: I thought about that, too. But then, they are currently not used in any FFT context as far as I could see.

`T_MINUS_ONE_DIV_TWO` is only used in the sqrt implementation and I couldn't find any usage of `T` at all.

algebra-core/src/fields/mod.rs
paberr: Yes, at least if there is a small subgroup defined it has a different semantic. I'll update the comment.

algebra-core/src/fields/mod.rs
paberr: So, `SMALL_SUBGROUP_BASE_ADICITY`?

Pratyush: Sure, that sounds good!

algebra-core/src/fields/mod.rs
Pratyush: Alright, that's fair. We can always move it in the future if the need arises. Thanks!

ff-fft/src/domain/general.rs
```diff
@@ -0,0 +1,220 @@
+//! This module contains a `GeneralEvaluationDomain` for
+//! performing various kinds of polynomial arithmetic on top of
+//! the scalar field.
+//!
+//! It is a wrapper around specific implementations of `EvaluationDomain` and allows
+//! to choose the most efficient implementation depending on the number of coefficients.
```
Pratyush:
```suggestion
//! a FFT-friendly finite field.
//!
//! It is a wrapper around specific implementations of `EvaluationDomain` that 
//! automatically chooses the most efficient implementation 
//! depending on the number of coefficients and the two-adicity of the prime.
```

ff-fft/src/domain/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+//! fields that are friendly to fast-fourier-transforms (FFTs).
+//!
+//! A field is FFT-friendly if it contains enough
+//! roots of unity to perform the FFT in O(n log n) time.
+//! These roots of unity comprise the domain over which
+//! polynomial arithmetic is performed.
+//!
+//! If the 2-adicity of the field is too small, but a small subgroup
+//! over the field is defined, we try to build a mixed-radix evaluation
+//! domain.
```
Pratyush: I feel like this doesn't belong here anymore

```suggestion
```

ff-fft/src/domain/radix2.rs
```diff
@@ -0,0 +1,312 @@
...
+        let log_size_of_group = size.trailing_zeros();
+
+        // libfqfft uses > https://github.com/scipr-lab/libfqfft/blob/e0183b2cef7d4c5deb21a6eaf3fe3b586d738fe0/libfqfft/evaluation_domain/domains/basic_radix2_domain.tcc#L33
+        if log_size_of_group > F::FftParams::TWO_ADICITY {
+            return None;
+        }
+
+        // Compute the generator for the multiplicative subgroup.
+        // It should be 2^(log_size_of_group) root of unity.
+        let group_gen = F::get_root_of_unity(size as usize)?;
```
Pratyush, line 61: Maybe we want to `debug_assert!` that this is indeed a root of unity for the given domain? (Also maybe for  `MixedRadixEvaluationDomain`)

ff-fft/src/domain/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+        let i = self
+            .evaluate_vanishing_polynomial(F::multiplicative_generator())
+            .inverse()
+            .unwrap();
+
+        cfg_iter_mut!(evals).for_each(|eval| *eval *= &i);
+    }
+
+    /// Given an index which assumes the first elements of this domain are the
+    /// elements of another (sub)domain with size size_s,
```
Pratyush:
```suggestion
    /// elements of another (sub)domain,
```

Pratyush:
Architectural question: Is `EvaluationDomain` supposed to work for arbitrary FFT domains (eg: in the future we might want to support arithmetic or geometric progression based domains, like in libfqfft), or is it intended to be limited to multiplicative subgroups?

If it is the former, then there are some functions (like `reindex_by_subdomain`) that currently have a default implementation, but this impl might be incorrect for domains that aren't multiplicative subgroups.

paberr:
> Architectural question: Is `EvaluationDomain` supposed to work for arbitrary FFT domains (eg: in the future we might want to support arithmetic or geometric progression based domains, like in libfqfft), or is it intended to be limited to multiplicative subgroups?

Right now it is intended to be limited to multiplicative subgroups.
> If it is the former, then there are some functions (like `reindex_by_subdomain`) that currently have a default implementation, but this impl might be incorrect for domains that aren't multiplicative subgroups.

True, but I didn't want to overcomplicate it at this stage. If we see need for other FFTs not based on multiplicative subgroups, we could still take care of this by e.g. having another trait.

Btw. I have an implementation of the geometric FFT in one of my branches, but it is just awfully slow and not really usable for any practical scenario. That's one of the reasons why I limited the scope to the multiplicative subgroups for now. :)

Pratyush:
Ok, that sounds good to me! Btw, in that case do we need a trait at all? Does the `enum` suffice?

paberr:
> Ok, that sounds good to me! Btw, in that case do we need a trait at all? Does the `enum` suffice?

Sure, an enum would work as well.
I addressed the remaining comments, added a few tests for the `get_root_of_unity` and also rebased to the current master.
*EDIT:* Forgot to check compilation of ff-fft. Fixing it right now. :)

ff-fft/src/domain/radix2.rs
paberr: I added it as well as actual tests for it.

Pratyush:
Ok this looks great! I'll merge once CI passes. Thanks for the patience and the awesome PR!

paberr:
> Ok this looks great! I'll merge once CI passes. Thanks for the patience and the awesome PR!

There seems to be some issue on nightly `error: linking with `cc` failed: exit code: 1`. I'm investigating that now, but it's not very obvious to me what's happening there.

Pratyush:
Hmm it seems to be happening on other PRs as well, so I don't think it's an issue  specific to this PR. I'm fine with merging as is.

paberr:
I also just realised that. It seems to be a problem with the current nightly.
I guess you can go ahead with merging then.

jon-chuang:
Hi @paberr ,
Could I ask if in your experiments, choosing a mixed evaluation domain could be more efficient even if the circuit does not cross the max two addicity?

For instance, for a circuit of size 5, choosing a circuit of size 2*3=6 requires less zero padding than choosing 2^4=8. 

This involves an optimisation problem to choose the smallest circuit from the available addicities. This can be phrased as a mixed-integer programming in `x_i`s with objective `min_{x_i} sum_{q_i primes} x_i log_2 q_i` with constraint `sum_{q_i primes} x_i log_2 q_i > log_2 n` If done right, this can reduce padding costs by close to 1/4 in expectation.

paberr:
Yes, I was thinking about it when implementing the mixed evaluation domain.
In the end, I didn't test it as MNT6 was the only curve for which we had a small subgroup defined.

But I think it is easy to adapt the benchmarks [here](https://github.com/scipr-lab/zexe/blob/master/ff-fft-benches/benches/fft.rs) to try it out with the MNT6 curve.
I can probably have a look at it on the weekend.

PR #181, opened 2020-04-11
feat(fields): Move `from_random_bytes` to Field

gakonst:
Also implement it for all extension fields.

### DIFF
--- a/algebra-core/algebra-core-derive/src/lib.rs
+++ b/algebra-core/algebra-core-derive/src/lib.rs
@@ -35,7 +35,7 @@ fn impl_serialize_field(
                 );
                 idents.pop();
             }
-        },
+        }
         _ => {
             serialize_body
                 .push(quote! { CanonicalSerialize::serialize(&self.#(#idents).*, writer)?; });
@@ -47,7 +47,7 @@ fn impl_serialize_field(
             uncompressed_size_body.push(
                 quote! { size += CanonicalSerialize::uncompressed_size(&self.#(#idents).*); },
             );
-        },
+        }
     }
 }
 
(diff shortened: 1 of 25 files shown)


kobigurk:
This is for usability, but I think it's worth clarifying - I think this method isn't expected to give the same results as in deserialization. This parses the number as being in a Montgomery representation, right? I guess generally this method isn't expected to match deserialization. @Pratyush ?

Pratyush:
Oh didn't realize you commented here, @kobigurk.

Yeah, the output of `from_random_bytes` doesn't match the output of calling deserialize. This is mostly to avoid having to do a multiplication by `R` in the process, but if that's not an issue, then one can change it to just call `deserialize` directly.

gakonst:
@Pratyush this is good for review now, Kobi got things working in https://github.com/gakonst/zexe/pull/1

Pratyush:
Hmm quick question: what's the difference between calling deserialize and calling `from_random_bytes`?

kobigurk:
> Hmm quick question: what's the difference between calling deserialize and calling `from_random_bytes`?

`from_random_bytes` removes all the extra bits after the modulus bits in each Fp component, where deserialize would just fail (or just return 0 now, actually). Additionally, deserialize for group elements happily accepts an infinity bit, where here we require we only get it in one out of the possible choices.

So this method has a high probability of succeeding and choosing uniformly an element out of the field, where deserialize doesn't.

Pratyush:
Hmm then to clarify the difference maybe we should add some documentation to the functions? Also to point out that these are intended mostly for sampling random field/group elements from hash-function/RNG output.

gakonst:
@Pratyush wdyt https://github.com/scipr-lab/zexe/pull/181/commits/3ff86c7093ab7fa3207c52bfc6e5ca284256adf7?

Pratyush:
Looks great! btw do you think we should move the functions on `GroupAffine` to the `AffineCurve` trait? Or do you think it's fine as is?

gakonst:
Good point, https://github.com/scipr-lab/zexe/pull/181/commits/e439be9244dbb2ad13ff75c18359a0c033c0cad2

Pratyush:
Awesome, thanks!

PR #189, opened 2020-04-14
Adds alloc_constant to AllocGadget

kobigurk:
This provides a uniform method to use a constant value in the place of booleans, uints, field elements and group elements.

### DIFF
--- a/crypto-primitives/src/commitment/pedersen/constraints.rs
+++ b/crypto-primitives/src/commitment/pedersen/constraints.rs
@@ -98,7 +98,10 @@ where
     W: PedersenWindow,
     ConstraintF: PrimeField,
 {
-    fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(cs: CS, val: T) -> Result<Self, SynthesisError>
+    fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(
+        cs: CS,
+        val: T,
+    ) -> Result<Self, SynthesisError>
     where
         T: Borrow<PedersenParameters<G>>,
     {
@@ -149,7 +152,10 @@ where
     G: Group,
     ConstraintF: PrimeField,
 {
-    fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(cs: CS, val: T) -> Result<Self, SynthesisError>
+    fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(
+        cs: CS,
+        val: T,
+    ) -> Result<Self, SynthesisError>
     where
         T: Borrow<PedersenRandomness<G>>,
     {
--- a/crypto-primitives/src/commitment/pedersen/constraints.rs
+++ b/crypto-primitives/src/commitment/pedersen/constraints.rs
@@ -98,6 +98,13 @@ where
     W: PedersenWindow,
     ConstraintF: PrimeField,
 {
+    fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(cs: CS, val: T) -> Result<Self, SynthesisError>
+    where
+        T: Borrow<PedersenParameters<G>>,
+    {
+        Self::alloc(cs, || Ok(val))
+    }
+
     fn alloc<F, T, CS: ConstraintSystem<ConstraintF>>(
         _cs: CS,
         value_gen: F,
@@ -142,6 +149,13 @@ where
     G: Group,
     ConstraintF: PrimeField,
 {
+    fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(cs: CS, val: T) -> Result<Self, SynthesisError>
+    where
+        T: Borrow<PedersenRandomness<G>>,
+    {
+        Self::alloc(cs, || Ok(val))
+    }
+
     fn alloc<F, T, CS: ConstraintSystem<ConstraintF>>(
         cs: CS,
         value_gen: F,
--- a/crypto-primitives/src/commitment/pedersen/constraints.rs
+++ b/crypto-primitives/src/commitment/pedersen/constraints.rs
@@ -99,25 +99,13 @@ where
     ConstraintF: PrimeField,
 {
     fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(
-        cs: CS,
+        _cs: CS,
         val: T,
     ) -> Result<Self, SynthesisError>
     where
         T: Borrow<PedersenParameters<G>>,
     {
-        Self::alloc(cs, || Ok(val))
-    }
-
-    fn alloc<F, T, CS: ConstraintSystem<ConstraintF>>(
-        _cs: CS,
-        value_gen: F,
-    ) -> Result<Self, SynthesisError>
-    where
-        F: FnOnce() -> Result<T, SynthesisError>,
-        T: Borrow<PedersenParameters<G>>,
-    {
-        let temp = value_gen()?;
-        let parameters = temp.borrow().clone();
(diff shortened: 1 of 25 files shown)


paberr:
This is great! We use [a very similar approach](https://github.com/nimiq/nano-sync/blob/master/src/gadgets/alloc_constant.rs) in our own code, but only implemented it for the types we needed.
Having it as part of the Zexe library would be amazing!

gakonst:
Extra context:

```
Kobi Gurkan, [15.04.20 11:57]
[In reply to Georgios Konstantopoulos]
alloc doesn’t enforce that the value that it’s in the circuit is what we alloced

Kobi Gurkan, [15.04.20 11:57]
meaning, if at one point I allocated G2 generator and I didn’t constrain it to be that element, then the prover can cheat and put w/e they want

Kobi Gurkan, [15.04.20 11:58]
I should have really used constants rather than allocating new variables - G2 generator is composed of two Fp2 coordinatoes, each of which are composed of an Fp element which is a constant (coefficient times the variable One)
```

crypto-primitives/src/commitment/pedersen/constraints.rs
```diff
@@ -98,6 +98,16 @@ where
...
 {
+    fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(
+        cs: CS,
+        val: T,
+    ) -> Result<Self, SynthesisError>
+    where
+        T: Borrow<PedersenParameters<G>>,
+    {
+        Self::alloc(cs, || Ok(val))
+    }
```
Pratyush: This is okay because there's nothing to allocate, right?

crypto-primitives/src/commitment/pedersen/constraints.rs
```diff
@@ -142,6 +152,16 @@ where
...
     ConstraintF: PrimeField,
 {
+    fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(
+        cs: CS,
+        val: T,
+    ) -> Result<Self, SynthesisError>
+    where
+        T: Borrow<PedersenRandomness<G>>,
+    {
+        Self::alloc(cs, || Ok(val))
```
Pratyush: Should this call `alloc_constant` of the underlying bytes? Maybe something like:
```suggestion
		let temp = value_gen()?;
        let mut result = vec![];
        for (i, byte) in to_bytes![temp.borrow().0].unwrap().into_iter().enumerate() {
        	let cur = UInt8::alloc_constant(cs.ns(|| format!("byte {}", i)), || Ok(byte))?;
        	result_bytes.push(cur);
        }
        Ok(PedersenRandomnessGadget(result))
```

Pratyush: (Also similarly for the Bowe-Hopwood-Pedersen hash, right?

r1cs-std/src/bits/boolean.rs
```diff
@@ -235,6 +235,16 @@ impl PartialEq for AllocatedBit {
...
 
 impl<ConstraintF: Field> AllocGadget<bool, ConstraintF> for AllocatedBit {
+    fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(
+        _cs: CS,
+        _t: T,
+    ) -> Result<Self, SynthesisError>
+    where
+        T: Borrow<bool>,
+    {
+        unimplemented!();
```
Pratyush, line 245: Do we want to implement this via the `CS::one()` idea?

crypto-primitives/src/commitment/pedersen/constraints.rs
kobigurk: Exactly, that was the thought.
Maybe we should reverse the order though - move the code to alloc_constant and have alloc call that, for clarity :)

crypto-primitives/src/commitment/pedersen/constraints.rs
kobigurk: You're right, yeah!

r1cs-std/src/bits/boolean.rs
kobigurk: Hmm, good point. We can just convert the bool to a bit and multiply by CS::one.

r1cs-std/src/fields/fp/mod.rs
```diff
@@ -584,6 +584,14 @@ impl<F: PrimeField> Clone for FpGadget<F> {
 }
 
 impl<F: PrimeField> AllocGadget<F, F> for FpGadget<F> {
+    #[inline]
+    fn alloc_constant<T, CS: ConstraintSystem<F>>(mut cs: CS, t: T) -> Result<Self, SynthesisError>
+    where
+        T: Borrow<F>,
+    {
+        Self::zero(cs.ns(|| "zero"))?.add_constant(cs.ns(|| "add constant"), t.borrow())
```
Pratyush: Underneath the hood it gets reduced to something like `LinearCombination::from((constant, CS::one())`, so it might clearer to directly say that

r1cs-std/src/fields/fp12.rs
```diff
@@ -834,6 +834,17 @@ where
...
 {
+    #[inline]
+    fn alloc_constant<T, CS: ConstraintSystem<ConstraintF>>(
+        mut cs: CS,
+        t: T,
+    ) -> Result<Self, SynthesisError>
+    where
+        T: Borrow<Fp12<P>>,
+    {
+        Self::zero(cs.ns(|| "zero"))?.add_constant(cs.ns(|| "add constant"), t.borrow())
```
Pratyush, line 845: Is there a reason not to just allocate each coordinate individually, besides this just being shorter?

kobigurk: Just wanted to benefit from the `add_constant` convenience, it should be equivalent. Wdyt?

r1cs-std/src/fields/fp/mod.rs
kobigurk: True. I can change it to that.

crypto-primitives/src/commitment/pedersen/constraints.rs
Pratyush: Yeah, that's probably better =)

r1cs-std/src/fields/fp12.rs
Pratyush: Sure, sounds good to me =)

crypto-primitives/src/commitment/pedersen/constraints.rs
kobigurk: I think this comment applies only to the Pedersen commitment, and the first comment to both Pedersen and BH hashes.

r1cs-std/src/bits/boolean.rs
kobigurk: This is going to be a bigger change than expected - it requires changing the `variable` field to a `ConstraintVar` which requires a type parameter, which will affect consumers. Any clever solution?

Pratyush: hmm good point. I'll address this in the WIP constraint system refactor

Pratyush: (Which will require adding the type parameter anyway)

kobigurk: Alright :)

kobigurk:
@Pratyush I handled all the comments besides the AllocatedBit one (which I believe shouldn't hurt anyone at the moment, as they'd be using Boolean::constant anyway).

PR #198, opened 2020-04-27
Merkle tree fixes

swasilyev:
Noticed two issues with the Merkle tree implementation
1. height = log(number_of_nodes) resulted in height = 0 for number_of_nodes = 1
2. It added an additional layer (a dummy node) even when effective tree height matched the required height, thus resulting in wrong Merkle root


Pratyush:
Great, thanks! Can you ensure that these changes are consistent with the gadget as well?

Pratyush:
ping @swasilyev, have you had a chance to look check whether these changes are consistent with the merkle tree?

swasilyev:
Sorry, forgot about it. Both the fixes deal with tree generation, not the Merkle proof verification (i guess this is what gadget does), so it should be ok. Let me fix the CI.

Pratyush:
Awesome thanks!

PR #199, opened 2020-04-28
allow log_size_of_group == Self::FftParams::TWO_ADICITY

weikengchen:
@paberr I feel here we should as well allow `log_size_of_group == Self::FftParams::TWO_ADICITY`.

Any thoughts?


paberr:
You're completely right. I must have missed that.
Thanks for spotting that and creating a patch!

weikengchen:
Great!

I discovered that due to an attempt of cycling of curves with MNT6-298, which, you know, has a small two-adicity.

PR #200, opened 2020-05-04
fixed some issues causing asm not to be compiled

jon-chuang:


Pratyush:
Oops!

algebra-core/src/lib.rs
```diff
@@ -6,13 +6,13 @@
 #![deny(unused_comparisons, bare_trait_objects, const_err, unused_must_use)]
 #![deny(unused_mut, unused_unsafe, private_in_public)]
 #![cfg_attr(
-    all(
+    not(all(
```
Pratyush: Hmm why do we want to negate here? `unsafe` is only required atm when all of these conditions are satisfied, right?

Pratyush:
Ok I pushed a fix which should make it always compile. I also consolidated the conditional logic in `build.rs` to make it easier to audit. Thanks for bringing this to my attention!

jon-chuang:
Ok, ran the bench, everything seems to be in order.

PR #217, opened 2020-06-12
added Edwards curves for MNT-298 and MNT-753

weikengchen:
With the help of Pratyush, we found Edwards curve parameters for MNT-298 and MNT-753. The former was easy. The latter took ~$1k AWS credits together with half a month of two machines in RISELab.

Comments are made extensive to aid others in case they need to construct an Edwards curve and want to know how to properly write the parameters in Zexe.

PR #218, opened 2020-06-13
Add Edwards curves over MNT-298 and MNT-753

weikengchen:
With the help of Pratyush, we found Edwards curve parameters for MNT-298 and MNT-753. The former was easy. The latter took ~$1k AWS credits together with half a month of two machines in RISELab.

Comments are made extensive to aid others in case they need to construct an Edwards curve and want to know how to properly write the parameters in Zexe.

This is, in spirit, similar to https://github.com/scipr-lab/zexe/pull/217


weikengchen:
wait for CI

PR #225, opened 2020-06-18
Implement BN curves

yelhousni:
This PR implements infrastructure for BN family of curves and instantiate [BN256 from Etherem](https://github.com/ethereum/go-ethereum/tree/master/crypto/bn256) (EDIT: curve renamed to BN254).
*Type: Feature
Label: Ready to review
Priority: Medium*

## Motivation
Support of BN family that is widely used for pairing-based cryptography (Ethereum, libsnark, DIZK, Halo).

## Followups
- [ ] Instantiate [BN254b from DIZK](https://github.com/scipr-lab/dizk/tree/8381582251337f3cd0b770f87251ad47719fae46/src/main/java/algebra/curves/barreto_naehrig/bn254b)
- [ ] Instantiate [BN382 from o1-labs](https://github.com/o1-labs/zexe/tree/master/algebra/src/bn_382) for [Halo-like recursion](https://eprint.iacr.org/2019/1021.pdf)


yelhousni:
@Pratyush @howardwu @kobigurk That would be great if you can review this!

algebra-core/src/curves/models/bn/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+
+        f
+    }
+
+    fn final_exponentiation(f: &Self::Fqk) -> Option<Self::Fqk> {
+        /*
+          Easy part: result = elt^((q^6-1)*(q^2+1)).
+          Follows, e.g., Beuchat et al page 9, by computing result as follows:
+             elt^((q^6-1)*(q^2+1)) = (conj(elt) * elt^(-1))^(q^2+1)
+        */
```
Pratyush: Just for consistency with the rest of the library

```suggestion
        // Easy part: result = elt^((q^6-1)*(q^2+1)).
        // Follows, e.g., Beuchat et al page 9, by computing result as follows:
        //   elt^((q^6-1)*(q^2+1)) = (conj(elt) * elt^(-1))^(q^2+1)
```

algebra-core/src/curves/models/bn/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+                  by computing:
+
+                  result = elt^(q^3 * (12*z^3 + 6z^2 + 4z - 1) +
+                                q^2 * (12*z^3 + 6z^2 + 6z) +
+                                q   * (12*z^3 + 6z^2 + 4z) +
+                                1   * (12*z^3 + 12z^2 + 6z + 1))
+                  which equals
+
+                  result = elt^( 2z * ( 6z^2 + 3z + 1 ) * (q^4 - q^2 + 1)/r ).
+                */
```
Pratyush: Ditto

```suggestion
                // Hard part follows Laura Fuentes-Castaneda et al. "Faster hashing to G2"
                // by computing:
                //
                // result = elt^(q^3 * (12*z^3 + 6z^2 + 4z - 1) +
                //               q^2 * (12*z^3 + 6z^2 + 6z) +
                //               q   * (12*z^3 + 6z^2 + 4z) +
                //               1   * (12*z^3 + 12z^2 + 6z + 1))
                // which equals
                //
                // result = elt^( 2z * ( 6z^2 + 3z + 1 ) * (q^4 - q^2 + 1)/r ).
```

algebra/Cargo.toml
```diff
@@ -32,11 +32,12 @@ rand_xorshift = "0.2"
 default = [ "std" ]
 
 full = [
-  "bls12_377", "bls12_381", "cp6_782", "bw6_761", "mnt4_298", "mnt4_753",
+  "bn256", "bls12_377", "bls12_381", "cp6_782", "bw6_761", "mnt4_298", "mnt4_753",
   "mnt6_298", "mnt6_753", "ed_on_bls12_377", "ed_on_cp6_782",
   "ed_on_bw6_761", "ed_on_bls12_381", "ed_on_mnt4_298", "ed_on_mnt4_753"
 ]
 
+bn256 = []
```
Pratyush:
```suggestion
  "bn254", "bls12_377", "bls12_381", "cp6_782", "bw6_761", "mnt4_298", "mnt4_753",
  "mnt6_298", "mnt6_753", "ed_on_bls12_377", "ed_on_cp6_782",
  "ed_on_bw6_761", "ed_on_bls12_381", "ed_on_mnt4_298", "ed_on_mnt4_753"
]

bn254 = []
```

algebra/src/bn256/curves/g2.rs
```diff
@@ -0,0 +1,114 @@
...
+        field_new!(Fq, BigInteger256([
+            0x38e7ecccd1dcff67,
+            0x65f0b37d93ce0d3e,
+            0xd749d0dd22ac00aa,
+            0x0141b9ce4a688d4d,
+        ])),
+    );
+
+    /// COFACTOR =
+    #[rustfmt::skip]
```
Pratyush: Do we know what the cofactor is?

algebra/src/bn256/fields/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+#[cfg(feature = "bn256")]
+pub use self::fq6::*;
+
+#[cfg(feature = "bn256")]
+pub mod fq12;
+#[cfg(feature = "bn256")]
+pub use self::fq12::*;
+
+#[cfg(all(feature = "bn256", test))]
+mod tests;
```
Pratyush:
```suggestion
#[cfg(feature = "bn254")]
pub mod fr;
#[cfg(feature = "bn254")]
pub use self::fr::*;

#[cfg(feature = "bn254")]
pub mod fq;
#[cfg(feature = "bn254")]
pub use self::fq::*;

#[cfg(feature = "bn254")]
pub mod fq2;
#[cfg(feature = "bn254")]
pub use self::fq2::*;

#[cfg(feature = "bn254")]
pub mod fq6;
#[cfg(feature = "bn254")]
pub use self::fq6::*;

#[cfg(feature = "bn254")]
pub mod fq12;
#[cfg(feature = "bn254")]
pub use self::fq12::*;

#[cfg(all(feature = "bn254", test))]
mod tests;
```

algebra/src/bn256/mod.rs
```diff
@@ -0,0 +1,25 @@
+//! This module implements the BN256 curve used in [[go-ethereum]](https://github.com/ethereum/go-ethereum/tree/master/crypto/bn256/cloudflare).
+//! The name denotes that it is a Barreto--Naehrig curve of embedding degree 12,
+//! defined over a 254-bit (prime) field. The scalar field is highly 2-adic.
+//! **CAUTION: This curve does not satisfy the 128-bit security level anymore.**
```
Pratyush:
```suggestion
//! This module implements the BN254 curve that was sampled as part of the [[BCTV14]][https://eprint.iacr.org/2013/879.pdf] paper .
//! The name denotes that it is a Barreto--Naehrig curve of embedding degree 12,
//! defined over a 254-bit (prime) field. The scalar field is highly 2-adic.
//!
//! This curve is also implemented in [libff](https://github.com/scipr-lab/libff/tree/master/libff/algebra/curves/alt_bn128).
//! It is the same as the `bn256` curve used in Ethereum (eg: [go-ethereum](https://github.com/ethereum/go-ethereum/tree/master/crypto/bn256/cloudflare)).
//!
//! #CAUTION 
//! **This curve does not satisfy the 128-bit security level anymore.**
```

algebra/src/lib.rs
```diff
@@ -32,6 +32,13 @@ pub use std::{boxed::Box, format, vec, vec::Vec};
 
 pub use algebra_core::*;
 
+///////////////////////////////////////////////////////////////////////////////
+#[cfg(feature = "bn256")]
+pub mod bn256;
+#[cfg(feature = "bn256")]
+pub use bn256::Bn256;
```
Pratyush:
```suggestion
#[cfg(feature = "bn254")]
pub mod bn254;
#[cfg(feature = "bn254")]
pub use bn254::Bn254;
```

Pratyush (changes requested):
Looks great, just some naming suggestions (`bn256` -> `bn254`).

(Should be accompanied by renaming the file in the file-system)

algebra-core/src/curves/models/bn/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    fn final_exponentiation(f: &Self::Fqk) -> Option<Self::Fqk> {
+        // Easy part: result = elt^((q^6-1)*(q^2+1)).
+        // Follows, e.g., Beuchat et al page 9, by computing result as follows:
+        //   elt^((q^6-1)*(q^2+1)) = (conj(elt) * elt^(-1))^(q^2+1)
+
+        // f1 = r.conjugate() = f^(p^6)
+        let mut f1 = *f;
+        f1.frobenius_map(6);
```
Pratyush:
```suggestion
        let mut f1 = *f;
        f1.conjugate();
```

algebra/Cargo.toml
yelhousni: > Looks great, just some naming suggestions (`bn256` -> `bn254`).

Yeah, I name it after Geth (https://github.com/ethereum/go-ethereum/tree/master/crypto/bn256) although I don't really agree with the name :p

algebra/src/bn256/curves/g2.rs
yelhousni: Yes, in sage:
```sage
q=21888242871839275222246405745257275088696311157297823662689037894645226208583
Fq=GF(q)
E=EllipticCurve(Fq,[0,3])
ZZ.<z>=Fq[]
Fq2.<u>=GF(q^2, modulus=z^2+1)
Et=EllipticCurve(Fq2,[0,3/(9+u)])
cofactor = Et.order() // E.order()
```

algebra-core/src/curves/models/bn/mod.rs
yelhousni: Right! `frobenius_map(6)` is a `conjugate` but I ported the BLS12 final exp easy part. We should change it there too.

algebra/Cargo.toml
Pratyush: Yeah, I don’t know why they renamed it, but in libsnark it’s bn254, so I think being consistent with that would be nice. (Also Zcash used it as the bn254 curve in the past)

kobigurk: If I remember correctly, the history goes like this:
* bn256 from cloudflare and google implement a BN curve that's unrelated to Zcash. When the Ethereum devs ported it to the libsnark curve, they kept the name.
* bn128 is the name that was used by libsnark and then by Zcash, denoting 128-bit security.
* Later on, Zcash and the community mostly decided to switch to the bn254 naming, since it wasn't 128-bit security anymore and 254-bit is the size of the curve.

I believe that the pairing-friendly curve draft from IETF actually calls this curve BN_SNARK1.

Pratyush:
Thanks for the helpful chronology! Given this info, what do you suggest we
should name the curve? I think bn256 doesn’t make sense if there are two
different curves with that name.

kobigurk:
IMO bn254.

Pratyush:
Ok sounds good then!

yelhousni:
Thank you Kobi for the chronology! If we are to choose the name that makes most sense, then `bn254` is the best option indeed. We just need to mention (in `mod.rs` for instance) that the curve is known also as `bn256` in Geth and `bn128` in libff.

algebra/src/bn256/curves/g2.rs
mratsim: The order on G2 for BN curves is `r∗(2p−r)` with `r` the G1 curve order, so the G2 cofactor is always (2p-r)

mratsim:
For naming I suggest BN254_Snarks as in the literature BN254 often (always?) refers to the one from Nogami et al with parameter `-(2^62 + 2^55 + 1)`

MCL also uses similar scheme to differentiate both: https://github.com/herumi/mcl#support-curves

algebra/src/bn256/curves/g2.rs
yelhousni: Yes indeed. Here is a good explanation on why: https://crypto.stackexchange.com/questions/64064/order-of-twisted-curve-in-pairings

Pratyush:
Hey @mratsim, thanks for the additional perspective!

I still feel that `bn254` is fine as it is well known in the context of SNARK systems (which this library is largely targeted towards). I also think that the literature moving forward will probably not use any BN254 curve, as these curves are  not secure. For the same reason, existing systems will probably eventually be deprecated the curve anyway.

Pratyush (approved):
LGTM!

PR #227, opened 2020-06-19
Add mul_by_inverse to FieldGadget

ValarDragon:
This implements `mul_by_inverse` to FieldGadget. This computes `a.mul_by_inverse(b) = a.mul(b.inverse())` but is computed in fewer constraints.

### DIFF
--- a/r1cs-std/src/fields/mod.rs
+++ b/r1cs-std/src/fields/mod.rs
@@ -200,6 +200,29 @@ pub trait FieldGadget<F: Field, ConstraintF: Field>:
         Ok(inverse)
     }
 
+    // Returns (self / denominator), but requires fewer constraints than
+    // self * denominator.inverse()
+    // It is up to the caller to ensure that denominator is non-zero,
+    // since in that case the result is unconstrained.
+    fn mul_by_inverse<CS: ConstraintSystem<ConstraintF>>(
+        &self,
+        mut cs: CS,
+        denominator: &Self,
+    ) -> Result<Self, SynthesisError> {
+        let denominator_inv_native = denominator
+            .get_value()
+            .and_then(|val| val.inverse())
+            .get()?;
+        let result_native = self.get_value().get()? * &denominator_inv_native;
+
+        let result = Self::alloc(&mut cs.ns(|| "alloc mul_by_inverse result"), || {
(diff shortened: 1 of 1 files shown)


r1cs-std/src/fields/mod.rs
```diff
@@ -430,13 +453,23 @@ pub(crate) mod tests {
 
         let a_inv = a.inverse(cs.ns(|| "a_inv")).unwrap();
         a_inv
-            .mul_equals(cs.ns(|| "check_equals"), &a, &one)
+            .mul_equals(cs.ns(|| "check a_inv * a = 1"), &a, &one)
```
ValarDragon, line 456: Went ahead and fixed this namespace in the inverse test to be descriptive, so as to avoid namespace collisions.

r1cs-std/src/fields/mod.rs
```diff
@@ -200,6 +200,29 @@ pub trait FieldGadget<F: Field, ConstraintF: Field>:
         Ok(inverse)
     }
 
+    // Returns (self / denominator), but requires fewer constraints than
+    // self * denominator.inverse()
+    // It is up to the caller to ensure that denominator is non-zero,
+    // since in that case the result is unconstrained.
```
Pratyush, line 206: I know there's a disclaimed in the comment, but if the denominator does end up being zero, then you can set the result to be zero, right? Which is unsound? Is there a way to combine a non-zero check for the denominator with the constraint itself?

```suggestion
    /// Returns (self / denominator), but requires fewer constraints than
    /// self * denominator.inverse()
    /// It is up to the caller to ensure that denominator is non-zero,
    /// since in that case the result is unconstrained.
```

ValarDragon: Err this comment is wrong, my bad. Its only unconstrained if the denominator and numerator are zero. (If just the numerator is zero, then result must be 0. If just the denominator is 0, then the constraint can't be satisfied) 

But 0/0 is undefined in general. I know of no way to check that both the numerator and denominator are both non-zero without additional constraints.

Pratyush: Hmm so this isn't any different from the `inverse` operation that's already in the system. Ok, looks good me =)

PR #234, opened 2020-07-05
Re-add `From<uXX>` trait bounds on `PrimeField`

Pratyush:


Pratyush:
These bounds now live directly on `Field`, so I'm closing this for now.

PR #236, opened 2020-07-08
feat(serialization): add missing impls

gakonst:
Adds a few more serialization implementations which allow more complex structs to be serialized, such as the KeypairAssembly.

### DIFF
--- a/algebra-core/src/serialize/mod.rs
+++ b/algebra-core/src/serialize/mod.rs
@@ -11,7 +11,7 @@ pub use flags::*;
 #[doc(hidden)]
 pub use algebra_core_derive::*;
 
-use crate::{Cow, Vec};
+use crate::{Cow, ToOwned, Vec};
 
 /// Serializer in little endian format allowing to encode flags.
 pub trait CanonicalSerializeWithFlags: CanonicalSerialize {
--- a/algebra-core/src/serialize/mod.rs
+++ b/algebra-core/src/serialize/mod.rs
@@ -118,7 +118,7 @@ macro_rules! impl_uint {
         }
 
         impl ConstantSerializedSize for $ty {
-            const SERIALIZED_SIZE: usize = std::mem::size_of::<$ty>();
+            const SERIALIZED_SIZE: usize = core::mem::size_of::<$ty>();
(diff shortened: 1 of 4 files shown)


algebra-core/src/serialize/mod.rs
```diff
@@ -100,32 +103,42 @@ pub trait CanonicalDeserialize: Sized {
...
+            }
+        }
+    };
 }
 
+impl_uint!(u8);
+impl_uint!(u16);
+impl_uint!(u32);
+impl_uint!(u64);
+impl_uint!(usize);
```
Pratyush: Hmm because this can change depending on architecture, we might not want to impl this (unless we serialize `usize`s as `u64`s always)

groth16/src/generator/generic.rs
```diff
@@ -35,6 +36,7 @@ where
 
 /// This is our assembly structure that we'll use to synthesize the
 /// circuit into a QAP.
+#[derive(Clone, Debug, CanonicalSerialize, CanonicalDeserialize)]
```
Pratyush, line 39: This is a data structure used to synthesize constraints, but doesn't contain the resulting public keys. I don't expect that this should ever be serialized to disk, right?

algebra-core/src/serialize/mod.rs
```diff
@@ -137,7 +137,32 @@ impl_uint!(u8);
...
+    const SERIALIZED_SIZE: usize = core::mem::size_of::<u64>();
+    const UNCOMPRESSED_SIZE: usize = Self::SERIALIZED_SIZE;
+}
+
+impl CanonicalDeserialize for usize {
+    #[inline]
+    fn deserialize<R: Read>(reader: &mut R) -> Result<Self, SerializationError> {
+        let mut bytes = [0u8; Self::SERIALIZED_SIZE];
+        reader.read_exact(&mut bytes)?;
+        Ok(u64::from_le_bytes(bytes) as usize)
```
Pratyush: Maybe you want to invoke a [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) impl here? So that this doesn't silently fail on 32-bit targets or something?

PR #243, opened 2020-07-23
Added g2_h formula and missing value in comment

AntoineRondelet:
For the record, see: https://eprint.iacr.org/2015/247.pdf proposition 1 section 3.3


Pratyush:
Thanks!

PR #244, opened 2020-07-24
Simplify field assembly register declaration

jon-chuang:


jon-chuang:
Doesn't work on stable for some reason.

Pratyush:
Isn't that just because assembly is nightly-only?

PR #245, opened 2020-07-25
fix log2 function

Vawheter:
Seems that there is a little bug in *log2(x: usize)* function in the algebra-core lib. When *x* is 2^i, the function returns i+1.


Pratyush:
Thanks for the PR and for bringing this issue up! I adapted your code and fixed this in #274.

PR #246, opened 2020-07-26
Implement `ToConstraintField` for `AffineCurve`

weikengchen:
This PR implements `ToConstraintField` for `AffineCurve`, which is to facilitate implementations of recursive proofs. To do so, `ToConstraintField` is now implemented for all fields, in which a field can be represented by the field's `BaseRepresentationField`, defined as follows:

- Fp320, Fp384, ...  -> Themselves
- Cubic of field X     -> X's BaseRepresentationField
- Quad of field X      -> X's BaseRepresentationField

A few repetitions of requirements are used to make the Rust compiler happy.

This PR has been used (and tested) in the ongoing implementation of Marlin's constraints.

Looking for initial feedback.

### DIFF
--- a/algebra-core/src/curves/mod.rs
+++ b/algebra-core/src/curves/mod.rs
@@ -2,7 +2,8 @@ use crate::{
     bytes::{FromBytes, ToBytes},
     fields::{Field, PrimeField, SquareRootField},
     groups::Group,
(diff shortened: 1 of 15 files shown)


weikengchen:
Ah, the repo from which I made the PR has been tainted. 

So I guess it is better for me to apply this to the refactored version instead. Closed for now.

PR #253, opened 2020-08-06
Add unchecked serialization/deserialization

Pratyush:
Add unchecked serialization and deserialization, for the case when one knows that the serialized output will be stored in a trusted location (for example: serialized and deserialized by the same machine).

cc @ChaitanyaKonda @iAmMichaelConnor, this might solve your use case in a cleaner way.

cc also @kobigurk @paberr

### DIFF
--- a/algebra-core/src/serialize/mod.rs
+++ b/algebra-core/src/serialize/mod.rs
@@ -52,6 +52,7 @@ pub trait ConstantSerializedSize: CanonicalSerialize {
 pub trait CanonicalSerialize {
     /// Serializes `self` into `writer`.
     fn serialize<W: Write>(&self, writer: &mut W) -> Result<(), SerializationError>;
+
     fn serialized_size(&self) -> usize;
 
     /// Serializes `self` into `writer` without compression.
@@ -59,6 +60,15 @@ pub trait CanonicalSerialize {
     fn serialize_uncompressed<W: Write>(&self, writer: &mut W) -> Result<(), SerializationError> {
         self.serialize(writer)
     }
+
+    /// Serializes `self` into `writer` without compression, and without performing
+    /// validity checks. Should be used *only* when there is no danger of
+    /// adversarial manipulation of the output.
+    #[inline]
+    fn serialize_unchecked<W: Write>(&self, writer: &mut W) -> Result<(), SerializationError> {
+        self.serialize_uncompressed(writer)
+    }
+
     #[inline]
     fn uncompressed_size(&self) -> usize {
         self.serialized_size()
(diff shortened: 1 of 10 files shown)


algebra-core/src/serialize/mod.rs
```diff
@@ -322,6 +375,20 @@ macro_rules! impl_prime_field_serializer {
...
+                reader: &mut R,
+            ) -> Result<Self, crate::serialize::SerializationError> {
+                const BYTE_SIZE: usize = $byte_size;
+
+                let (_, output_byte_size) =
+                    crate::serialize::buffer_bit_byte_size($field::<P>::size_in_bits());
+
+                let mut masked_bytes = [0; BYTE_SIZE];
+                reader.read_exact(&mut masked_bytes[..output_byte_size])?;
+                Ok(Self::read(&masked_bytes[..])?)
```
paberr: This is duplicate code with the `deserialize` just above.
The default implementation will do the job and the implementation for `deserialize_unchecked` can be removed.

algebra-core/src/serialize/mod.rs
```diff
@@ -284,6 +321,22 @@ macro_rules! impl_prime_field_serializer {
...
+                const BYTE_SIZE: usize = $byte_size;
+
+                let (_, output_byte_size) =
+                    crate::serialize::buffer_bit_byte_size($field::<P>::size_in_bits());
+
+                let mut bytes = [0u8; BYTE_SIZE];
+                self.write(&mut bytes[..])?;
+
+                writer.write_all(&bytes[..output_byte_size])?;
+                Ok(())
```
paberr: This code is basically doing the same as the `serialize` or calling `self.serialize_with_flags(writer, crate::serialize::EmptyFlags)`.
The only difference is the handling of the empty flags.
Did you implement this for performance reasons?

paberr (commented):
I agree this is probably the cleaner way to do it. 👍 
One remark though: The `_unchecked` variants are now incompatible with the `_uncompressed` methods for serialization/deserialization. The `_unchecked` implementations write the infinity bit explicitly into its own byte, while the `_uncompressed` code encodes it into the non-used bits.

Did you choose to use an explicit bool for performance reasons?
If yes, I think adding a note to the documentation would be good, indicating this difference.

algebra-core/src/serialize/mod.rs
Pratyush: Yeah, I didn't want the additional overhead from the flag conversion. But I guess I should benchmark and compare the difference.

Pratyush:
> Did you choose to use an explicit bool for performance reasons?

I did omit the flag handling for affine structs  for performance reasons, but I should check that it actually makes a difference by benchmarking.

kobigurk:
This is very useful! I actually might need something like this soon.
That said, I'm highly in favor of making it compatible with the normal decompression. @Pratyush do you think you're going to decide by benchmarking?

Pratyush:
@kobigurk @paberr I think I'll just make it compatible with normal decompression, because I suspect most of the overhead is due to subgroup checks, and not due to bit manipulation. (I'm writing benchmarks right now to verify this)

algebra-core/src/serialize/mod.rs
Pratyush: There was no difference =D

Pratyush:
> I suspect most of the overhead is due to subgroup checks, and not due to bit manipulation. 

Indeed, most of the overhead is due to the prime-group check.

PR #255, opened 2020-08-07
Improve Serialization/Deserialization API to use `mut reader/writer: R/W`

Pratyush:
In the process of working on #253, I decided to fix a small nit with the current API of `CanonicalSerialize/Deserialize`. *This is a breaking change*, albeit one that's easily fixed via the following substitutions: `s/\<reader\>: &mut R/mut reader: R/g` and `s/\<writer\>: &mut W/mut writer: W/g`, along with passing `reader`/`writer` by mutable reference to subroutines.

cc @kobigurk @paberr

### DIFF
--- a/algebra-benches/src/macros/ec.rs
+++ b/algebra-benches/src/macros/ec.rs
@@ -104,7 +104,7 @@ macro_rules! ec_bench {
             b.iter(|| {
                 count = (count + 1) % SAMPLES;
                 let index = count * num_bytes;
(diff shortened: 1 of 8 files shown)


kobigurk:
Looks like a good idea to me. I never saw what the best practice is btw. Is it written anywhere?

Pratyush:
I think here: https://rust-lang.github.io/api-guidelines/interoperability.html#c-rw-value

paberr:
The `mut reader/writer: R/W` change looks good to me. :+1:

PR #278, opened 2020-09-15
Add `ToConstraintFieldGadget`

weikengchen:
This PR adds `ToConstraintFieldGadget` to the refactored Zexe.

PR #284, opened 2020-09-17
ToConstraintFieldGadget for SW ProjectiveVar

weikengchen:
The previous PR https://github.com/scipr-lab/zexe/pull/278 does not include the case of SW `ProjectiveVar` (but only `AffineVar`), but it is used in the main implementations of `PairingVar` for SW curves.

So, when we are refactoring the Marlin gadgets, we discovered that `ToConstraintFieldGadget` for SW `ProjectiveVar` is needed. This PR provides so.

### DIFF
--- a/algebra-core/src/to_field_vec.rs
+++ b/algebra-core/src/to_field_vec.rs
@@ -23,6 +23,16 @@ impl<F: PrimeField> ToConstraintField<F> for F {
     }
 }
 
+impl<F: Field> ToConstraintField<F> for bool {
+    fn to_field_elements(&self) -> Result<Vec<F>, Error> {
+        if *self == true {
+            Ok(vec![F::one()])
+        }else{
+            Ok(vec![F::zero()])
+        }
+    }
+}
+
 // Impl for base field
 impl<F: Field> ToConstraintField<F> for [F] {
     #[inline]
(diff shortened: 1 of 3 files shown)


weikengchen:
The CI passes!

Pratyush:
Should this serialize also the zero bit? I.e., by converting to `SWAffine` and serializing that?

weikengchen:
You mean the `infinity` bit? 

The current upstream implementation only to_constraint_fields x and y for Affine for SW. Should we also include the infinity bool? I intentionally removed this one since some PR (I forgot) starts to omit the infinity bool.

weikengchen:
Ah, I misunderstood. So I double-read your comment and the code, now there are two action items:

1. Handle the case when `z=0`. Now it would simply fail, but ideally we should do `to_affine` like https://github.com/scipr-lab/zexe/blob/master/r1cs-std/src/groups/curves/short_weierstrass/mod.rs#L154 first, and convert it to filed gadgets from an affine.

2. Question: should we also to_constraint_field the infinity bit in SWAffine? Currently, it doesn't, but it seems safer to do so?

Pratyush:
yes to both!

weikengchen:
Done and CI passes. Please review it again. (The code is significantly shorter for ProjectiveVar!)

weikengchen:
Closed temporarily. (The branch has been wrongly used for some other changes).

Will later push one to the arkworks's.

Issue #283, opened 2020-09-17
to_bits_be?


weikengchen:
Our current `ToBitsGadget` already has `to_bits_le`,
```
/// Specifies constraints for conversion to a little-endian bit representation of `self`.
pub trait ToBitsGadget<F: Field> {
    /// Outputs the canonical little-endian bit-wise representation of `self`.
    ///
    /// This is the correct default for 99% of use cases.
    fn to_bits_le(&self) -> Result<Vec<Boolean<F>>, SynthesisError>;

    /// Outputs a possibly non-unique bit-wise representation of `self`.
    ///
    /// If you're not absolutely certain that your usecase can get away with a
    /// non-canonical representation, please use `self.to_bits()` instead.
    fn to_non_unique_bits_le(&self) -> Result<Vec<Boolean<F>>, SynthesisError> {
        self.to_bits_le()
    }
}
```

Would it be useful to also add `to_bits_be`, which would be getting the result of `to_bits_le` and then reversing it?
This is useful because legacy programs may have been using the old `to_bits`.

PR #285, opened 2020-09-18
to_bits_be

weikengchen:
About the issue https://github.com/scipr-lab/zexe/issues/283

If we decided that `to_bits_be` is useful, then this PR provides a straightforward implementation.


weikengchen:
CI passes.

PR #299, opened 2020-10-01
Allow creation of group element by raw X and Y coordinates

dark64:
This PR adds `from_xy_checked` function to `AffineCurve` trait which can be used to create a point from raw X and Y coordinates. This function returns a `Result` as on-curve check is performed.

We are in process of adding zexe as a backend to [ZoKrates](https://github.com/Zokrates/ZoKrates) project and this is currently a  block for us. As we work with raw x and y coordinates we need to have a way to construct associated `G1Affine` and `G2Affine` types from `PairingEngine` trait. This change would enables us to do so.

As a reference, this is supported for bellman in the [pairing_ce](https://docs.rs/pairing_ce/0.24.1/pairing_ce/trait.CurveAffine.html#tymethod.from_xy_checked) crate. Any feedback is appreciated on this.

### DIFF
--- a/algebra-core/src/curves/models/short_weierstrass_jacobian.rs
+++ b/algebra-core/src/curves/models/short_weierstrass_jacobian.rs
@@ -202,6 +202,22 @@ impl<P: Parameters> AffineCurve for GroupAffine<P> {
     fn mul_by_cofactor_inv(&self) -> Self {
         self.mul(P::COFACTOR_INV).into()
     }
+
+    fn from_xy_checked(x: Self::BaseField, y: Self::BaseField) -> Result<Self, ()> {
+        let infinity = x.is_zero() && y.is_zero();
+        let affine = Self {
+            x,
+            y,
+            infinity,
+            _params: PhantomData,
+        };
+
+        if !affine.is_on_curve() {
+            Err(())
+        } else {
+            Ok(affine)
+        }
+    }
 }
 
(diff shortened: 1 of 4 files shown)


Pratyush:
Do you think it makes sense to instead try to construct these by deserializing from the relevant bytes? I feel like that's less error prone and less likely to cause bugs.

algebra-core/src/curves/models/short_weierstrass_jacobian.rs
```diff
@@ -202,6 +202,22 @@ impl<P: Parameters> AffineCurve for GroupAffine<P> {
     fn mul_by_cofactor_inv(&self) -> Self {
         self.mul(P::COFACTOR_INV).into()
     }
+
+    fn from_xy_checked(x: Self::BaseField, y: Self::BaseField) -> Result<Self, ()> {
+        let infinity = x.is_zero() && y.is_zero();
```
Pratyush, line 207: (0, 0) is a point on some curves (eg: those of the form (`y^2 = x^3 + ax`)), so setting it to the point at infinity isn't correct.

dark64: oh right, didn't think of this :/

dark64:
I managed to do this using the `FromBytes` trait. I guess my proposed change is not needed after all, so I'm closing this. Thanks for the help! :+1:

Issue #305, opened 2020-10-14
Optimize MNT4/6 pairing (2-NAF Miller loop)


yelhousni:
Currently, MNT4/6 pairing is implemented in projective coordinate with a binary Miller loop. 
Note that the Hamming weight of `ATE_LOOP_COUNT` for both MNT298 and MNT753 is lower when decomposed in 2-NAF. I suggest to implement 2-NAF Miller loop for MNT4 and MNT6.

Issue #303, opened 2020-10-08
algebra-core fails to compile on nightly


tsunrise:
It seems that recent version of `algebra-core` cannot be compiled on nightly.
error looks like this
```
error[E0277]: cannot multiply-assign `&&<P as quadratic_extension::QuadExtParameters>::BaseField` to `<P as quadratic_extension::QuadExtParameters>::BaseField`
Error:    --> /home/runner/.cargo/git/checkouts/zexe-f74c42809fa6cfe1/85cac17/algebra-core/src/fields/models/quadratic_extension.rs:135:28
    |
135 |         self.c0.mul_assign(&element);
    |                            ^^^^^^^^ no implementation for `<P as quadratic_extension::QuadExtParameters>::BaseField *= &&<P as quadratic_extension::QuadExtParameters>::BaseField`
    |
    = help: the trait `std::ops::MulAssign<&&<P as quadratic_extension::QuadExtParameters>::BaseField>` is not implemented for `<P as quadratic_extension::QuadExtParameters>::BaseField`
help: consider further restricting the associated type
    |
134 |     pub fn mul_assign_by_basefield(&mut self, element: &P::BaseField) where <P as quadratic_extension::QuadExtParameters>::BaseField: std::ops::MulAssign<&&<P as quadratic_extension::QuadExtParameters>::BaseField> {
    |                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
here's the [error log](https://github.com/arkworks-rs/sumcheck/runs/1228605985) if you have access to it

jon-chuang:
nightly seems to be the problem as these type coercions are meant to be performed by rustc. We have some fixes in celo-org/zexe, but one can alternatively wait out the next nightly.

jgraef:
I bisected it to commit `08e2d4616613716362b4b49980ff303f2b9ae654` which is this [PR](https://github.com/rust-lang/rust/pull/73905). Looks like this is intended and won't be reversed.

PR #307, opened 2020-10-14
Optimize MNT4 pairing (2-NAF Miller loop)

yelhousni:
#305 
MNT4


jon-chuang:
Is there a reason why one does not use a bigger window size for the w-NAf?

yelhousni:
> Is there a reason why one does not use a bigger window size for the w-NAf?

The Miller loop iterates over the binary decomposition of `ATE_LOOP_COUNT` (double if 0, double-and-add if 1). Noting that point negation is cheap (x,y)->(x,-y), one can leverage a 2-NAF decomposition if it has more 0's than in the binary form.

jon-chuang:
Wouldn't it make sense to use a table here then?

yelhousni:
> Wouldn't it make sense to use a table here then?

@jon-chuang How would you suggest to do that? I am storing the pre-computed 2-NAF decomposition here https://github.com/scipr-lab/zexe/blob/48ccd3290ddf9502c005a44f0bc6f9e100c9ce8d/algebra/src/mnt4_753/curves/mod.rs#L40
and running a 1/-1 pattern matching in the Miller Loop here https://github.com/scipr-lab/zexe/blob/48ccd3290ddf9502c005a44f0bc6f9e100c9ce8d/algebra-core/src/curves/models/mnt4/mod.rs#L128

yelhousni:
For CI / Test (nightly) failing, see #303

Pratyush:
Hey @yelhousni, thanks for the PR!
We're transitioning the code in this repo over to https://github.com/arkworks-rs/algebra/; could you make the PR there instead?

Pratyush:
Hi @yelhousni do you mind re-opening this PR against arkworks-rs/algebra? Sorry for the inconvenience!

PR #309, opened 2020-10-15
replace frob(6) by unitary_inverse in bls12 in r1cs-std

yelhousni:
same as #231 but for `r1cs-std` crate


Pratyush:
Hey @yelhousni, thanks for the PR!
We're transitioning the code in this repo over to https://github.com/arkworks-rs/algebra/; could you make the PR there instead?

Pratyush:
I believe this PR was merged in arkworks-rs/algebra, right?

PR #331, opened 2020-12-20
Fix the `outline_lcs`

weikengchen:
After careful debugging, it appears that the current implementation of `outline_lcs` has two serious problems:

- It does not correctly create additional constraints. All the additional constraints should be `LC * 1 = V` where V is the new witness variable. In the end, all of them are `V * 1 = V`.

- It does not increase the count of `num_linear_combinations` correctly. So that in Marlin, when the next step wants to pad the matrices to be square, the new linear combinations go to the wrong linear combination entry, causing the outer check to fail.

This PR fixes these two problems and will close this issue in Marlin: https://github.com/arkworks-rs/marlin/issues/56

### DIFF
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -257,6 +257,105 @@ impl<F: Field> ConstraintSystem<F> {
         num_times_used
     }
 
+    /// Transform the map of linear combinations.
+    /// Specifically, allow the creation of additional witness assignments.
+    ///
+    /// This method is used as a subroutine of `inline_all_lcs` and `outline_lcs`.
+    ///
+    /// The transformer function is given references of this constraint system (&self),
+    /// number of times used, and the linear combination to be transformed.
+    ///     (&ConstraintSystem<F>, usize, &mut LinearCombination<F>)
+    ///
+    /// The transformer function returns numbers of new witness variables needed,
+    /// and a vector of new witness assignments (if not in the setup mode).
+    ///     (usize, Option<Vec<F>>)
+    ///
+    pub fn transform_lc_map(
+        &mut self,
+        transformer: &mut dyn FnMut(
+            &ConstraintSystem<F>,
+            usize,
+            &mut LinearCombination<F>,
+        ) -> (usize, Option<Vec<F>>),
+    ) {
+        // `transformed_lc_map` stores the transformed linear combinations.
+        let mut transformed_lc_map = BTreeMap::new();
+        let mut num_times_used = self.lc_num_times_used(false);
+
+        // This loop goes through all the LCs in the map, starting from
+        // the early ones. The transformer function is applied to the
+        // inlined LC, where new witness variables can be created.
+        for (&index, lc) in &self.lc_map {
+            let mut transformed_lc = LinearCombination::new();
+
+            // Inline the LC, unwrapping symbolic LCs that may constitute it,
+            // and updating them according to transformations in prior iterations.
+            for &(coeff, var) in lc.iter() {
+                if var.is_lc() {
+                    let lc_index = var.get_lc_index().expect("should be lc");
+
+                    // If `var` is a `SymbolicLc`, fetch the corresponding
+                    // inlined LC, and substitute it in.
+                    //
+                    // We have the guarantee that `lc_index` must exist in
+                    // `new_lc_map` since a LC can only depend on other
+                    // LCs with lower indices, which we have transformed.
+                    //
+                    let lc = transformed_lc_map
+                        .get(&lc_index)
+                        .expect("should be inlined");
+                    transformed_lc.extend((lc * coeff).0.into_iter());
+
+                    // Delete linear combinations that are no longer used.
+                    //
+                    // Deletion is safe for both outlining and inlining:
+                    // * Inlining: the LC is substituted directly into all use
+                    //   sites, and so once it is fully inlined, it is redundant.
+                    //
+                    // * Outlining: the LC is associated with a new variable `w`,
+                    //   and a new constraint of the form `lc_data * 1 = w`, where
+                    //   `lc_data` is the actual data in the linear combination.
+                    //   Furthermore, we replace its entry in `new_lc_map` with `(1, w)`.
+                    //   Once `w` is fully inlined, then we can delete the entry from `new_lc_map`
+                    //
+                    num_times_used[lc_index.0] -= 1;
+                    if num_times_used[lc_index.0] == 0 {
+                        // This lc is not used any more, so remove it.
+                        transformed_lc_map.remove(&lc_index);
+                    }
+                } else {
+                    // Otherwise, it's a concrete variable and so we
+                    // substitute it in directly.
+                    transformed_lc.push((coeff, var));
+                }
+            }
+            transformed_lc.compactify();
+
+            // Call the transformer.
+            let (num_new_witness_variables, new_witness_assignments) =
+                transformer(&self, num_times_used[index.0], &mut transformed_lc);
+
+            // Insert the transformed LC.
+            transformed_lc_map.insert(index, transformed_lc);
+
+            // Update the witness counter.
+            self.num_witness_variables += num_new_witness_variables;
+
+            // Supply additional witness assignments if not in the
+            // setup mode and if new witness variables are created.
+            if !self.is_in_setup_mode() && num_new_witness_variables > 0 {
+                assert!(new_witness_assignments.is_some());
+                if let Some(new_witness_assignments) = new_witness_assignments {
+                    assert_eq!(new_witness_assignments.len(), num_new_witness_variables);
+                    self.witness_assignment
+                        .extend_from_slice(&new_witness_assignments);
+                }
+            }
+        }
+        // Replace the LC map.
+        self.lc_map = transformed_lc_map;
+    }
+
     /// Naively inlines symbolic linear combinations into the linear
     /// combinations that use them.
     ///
@@ -266,34 +365,15 @@ impl<F: Field> ConstraintSystem<F> {
     /// do not contribute to the size of the multi-scalar multiplication, which
     /// is the dominating cost.
     pub fn inline_all_lcs(&mut self) {
-        let mut inlined_lcs = BTreeMap::new();
-        let mut num_times_used = self.lc_num_times_used(false);
-
-        for (&index, lc) in &self.lc_map {
-            let mut inlined_lc = LinearCombination::new();
-            for &(coeff, var) in lc.iter() {
-                if var.is_lc() {
-                    let lc_index = var.get_lc_index().expect("should be lc");
-                    // If `var` is a `SymbolicLc`, fetch the corresponding
-                    // inlined LC, and substitute it in.
-                    let lc = inlined_lcs.get(&lc_index).expect("should be inlined");
-                    inlined_lc.extend((lc * coeff).0.into_iter());
-
-                    num_times_used[lc_index.0] -= 1;
-                    if num_times_used[lc_index.0] == 0 {
-                        // This lc is not used any more, so remove it.
-                        inlined_lcs.remove(&lc_index);
-                    }
-                } else {
-                    // Otherwise, it's a concrete variable and so we
-                    // substitute it in directly.
-                    inlined_lc.push((coeff, var));
-                }
-            }
-            inlined_lc.compactify();
-            inlined_lcs.insert(index, inlined_lc);
+        // Only inline when a matrix representing R1CS is needed.
+        if !self.should_construct_matrices() {
+            return;
         }
-        self.lc_map = inlined_lcs;
+
+        // A dummy closure is used, which means that
+        // - it does not modify the inlined LC.
+        // - it does not add new witness variables.
+        self.transform_lc_map(&mut |_, _, _| (0, None));
     }
 
     /// If a `SymbolicLc` is used in more than one location and has sufficient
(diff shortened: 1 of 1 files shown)


weikengchen:
It turns out that I cannot assign a reviewer here. So, let me here cc Pratyush for a review. @Pratyush 

We may, in the future, has an improvement plan for `outline_lcs`, since the resulting saving is not as large as expected.

Pratyush:
Thanks for the PR! Given that there was a subtle bug in the logic here, would it be possible to add some comments documenting the algorithm?

weikengchen:
No problem. On the way!

weikengchen:
The documentation has been added. Please take a look :)

relations/src/r1cs/constraint_system.rs
```diff
@@ -334,19 +333,23 @@ impl<F: Field> ConstraintSystem<F> {
...
 
             if this_used_times * this_len > this_used_times + 2 + this_len {
-                should_promote = true;
+                should_outline = true;
             }
 
-            if should_promote {
+            // If outlining is worthwhile,
+            if should_outline {
+                // Add a new witness (the value of the linear combination).
```
Pratyush, line 429: Why not use `self.new_witness` here? It would avoid errors from manual management of witness variables.

relations/src/r1cs/constraint_system.rs
```diff
@@ -356,11 +359,11 @@ impl<F: Field> ConstraintSystem<F> {
                     self.witness_assignment.push(acc);
                 }
 
-                self.num_constraints += 1;
-
-                additional_a_constraints.push(index);
-                additional_c_constraints_witness_assignments.push(witness_assignment);
+                // Add a new constarint for this new witness
+                new_witness_linear_combinations.push(outlined_lc.clone());
+                new_witness_indices.push(witness_assignment);
```
Pratyush:
```suggestion
                // Add a new constraint for this new witness
                new_witness_linear_combinations.push(outlined_lc.clone());
                new_witness_indices.push(witness_assignment);
```

Also, why not use `self.constrain` here?

weikengchen:
For the `new_witness`, yes, good point, let me use `new_witness` instead.

For the new constraint via `enforce_constraint`, it has a problem---our code is modifying the lc_map but `enforce_constraint` would also push to it, causing a conflict (that our code may overwrite the change). But let me take a more careful look to see I can still use `enforce_constraint` in a black-box way if I save the `lc_map` earlier.

Pratyush:
For the `new_constraint` issue, maybe we can extract the common logic into a separate method that is called in both places?

weikengchen:
Yes. But I feel that probably I can call `new_constraint` directly if I do `self.lc_map = outlined_lcs;` earlier, since these new constraints all use outlined LCs (so no more processing is needed).

relations/src/r1cs/constraint_system.rs
```diff
@@ -240,12 +240,12 @@ impl<F: Field> ConstraintSystem<F> {
 
     /// Count the number of times a given LC is used within another LC
```
ValarDragon:
```suggestion
    /// Count the number of times each LC is used within other LCs in the constraint system
```

relations/src/r1cs/constraint_system.rs
```diff
@@ -240,12 +240,12 @@ impl<F: Field> ConstraintSystem<F> {
...
     fn lc_num_times_used(&self, count_sinks: bool) -> Vec<usize> {
-        // step 1: Identify all lcs that have been many times
         let mut num_times_used = vec![0; self.lc_map.len()];
 
         for (index, lc) in self.lc_map.iter() {
```
ValarDragon, line 246:
```suggestion
        // Iterate over every lc in constraint system
        for (index, lc) in self.lc_map.iter() {
```

weikengchen:
I use `new_constraint` directly, so the code is now slightly simplified.

I didn't use `new_witness` for a specific reason:
- The `new_witness` somehow has to be in the `for` loop of `self.lc_map`. Moving it out could lead to potential missing assignments.
- But `new_witness` takes `&mut self` as input, but this is forbidden (since `self` is currently used for iteration).
- If I want to allow `new_witness` to use `&mut self`, I need to clone `self.lc_map`, but the memory usage might be too high.

Thus, given that the logic of `new_witness` is simple, I just write the same code again. 

Note: I need to test the upstream Marlin for this PR to see if it is working. So don't merge now!

relations/src/r1cs/constraint_system.rs
ValarDragon: I know this wasn't part of this PR, but since this was being edited figured might as well improve the comments

relations/src/r1cs/constraint_system.rs
```diff
@@ -311,12 +311,11 @@ impl<F: Field> ConstraintSystem<F> {
 
         let used_times = self.lc_num_times_used(true);
```
ValarDragon: Can a comment be added here as to why sinks are being counted? Its not clear to me atm.

(Also the word sink is non-intuitive to me in this context, since we're not talking about a state machine / probabilistic process of any sort)

(We can make this into a new issue / PR though)

Pratyush:
Hmm in that case we should document that this is the same logic as `new_witness`.

Also, speaking of memory requirements, we should "garbage-collect" the LCs that are inlined/put in new variables, like we do in `inline_all_lcs`

relations/src/r1cs/constraint_system.rs
```diff
@@ -311,12 +311,11 @@ impl<F: Field> ConstraintSystem<F> {
...
-        let mut additional_a_constraints = Vec::<LcIndex>::new();
-        let mut additional_c_constraints_witness_assignments = Vec::<usize>::new();
+        let mut new_witness_linear_combinations = Vec::new();
+        let mut new_witness_indices = Vec::new();
 
+        // `outlined_lcs` stores the modified linear combinations.
+        let mut outlined_lcs = BTreeMap::new();
         for (&index, lc) in self.lc_map.iter() {
             let mut outlined_lc = LinearCombination::new();
             for &(coeff, var) in lc.iter() {
```
ValarDragon: Can we have a comment to describe what this for loop is doing? It looks `outlined_lcs` is acting as a map to store all prior replacements, and inlining all symbols.

Though I don't get why the core loop works though. (e.g. why does outlined_lcs always contain lc_index in the line `let lc = outlined_lcs.get(&lc_index).expect("should be inlined");`)

weikengchen:
Got it. Let me have some tries.

relations/src/r1cs/constraint_system.rs
```diff
@@ -334,19 +333,23 @@ impl<F: Field> ConstraintSystem<F> {
...
             outlined_lc.compactify();
 
-            let mut should_promote = false;
+            let mut should_outline = false;
 
+            // Check if outlining is worthwhile.
             let this_used_times = used_times[index.0];
             let this_len = outlined_lc.len();
 
             if this_used_times * this_len > this_used_times + 2 + this_len {
```
ValarDragon, line 423: Can we add a comment explaining the cost. (Namely that we change the arity from `this_len` repeated `this_used_times` to a constraint for `this_len * 1 = {new variable}` and then all of the `this_used_times` entries only have one usage.
```suggestion
            // Cost with no outlining = `lc_len * number of usages`
            // Cost with outlining is one constraint for `(lc_len) * 1 = {new variable}` and 
            // using that single new variable in each of the prior usages.
            // This has total cost `number_of_usages + lc_len + 2`
            if this_used_times * this_len > this_used_times + 2 + this_len {
```

ValarDragon (commented):
Added a couple of places for additional comments. None of these should block the PR though

ValarDragon:
Thanks for fixing this bug!

weikengchen:
More review is welcome. Let me merge the suggestions!

relations/src/r1cs/constraint_system.rs
weikengchen: Resolved by adding more explanations.

weikengchen:
I updated the suggestions. Currently testing with Marlin. 

@ValarDragon can do a final pass over the new comments that try to explain the loop.

relations/src/r1cs/constraint_system.rs
ValarDragon: Thanks! The algorithm now makes sense to me =)

relations/src/r1cs/constraint_system.rs
```diff
@@ -309,91 +310,111 @@ impl<F: Field> ConstraintSystem<F> {
...
                 if var.is_lc() {
                     let lc_index = var.get_lc_index().expect("should be lc");
+
                     // If `var` is a `SymbolicLc`, fetch the corresponding
                     // inlined LC, and substitute it in.
-                    let lc = outlined_lcs.get(&lc_index).expect("should be inlined");
-                    outlined_lc.extend((lc * coeff).0.into_iter());
+                    //
+                    // We have the guarantee that `lc_index` must exist in
+                    // `new_lc_msp` since a LC can only depend on other
```
Pratyush: msp -> map? (Here and elsewhere?)

weikengchen: Ah! Working on it.

I need to be careful. Recently my typo rate is very high.

weikengchen:
The upstream test passes, and it is ready to merge if no other issues pop up.

weikengchen:
(ps I need someone who has write access.)

Pratyush:
Doing a final review pass, and then I'll merge!

relations/src/r1cs/constraint_system.rs
```diff
@@ -309,91 +310,111 @@ impl<F: Field> ConstraintSystem<F> {
...
+                    // `new_lc_map` since a LC can only depend on other
+                    // LCs with lower indices, which we have modified.
+                    //
+                    let lc = new_lc_map.get(&lc_index).expect("should be inlined");
+                    inlined_lc.extend((lc * coeff).0.into_iter());
+
+                    num_times_used[lc_index.0] -= 1;
+                    if num_times_used[lc_index.0] == 0 {
+                        // This lc is not used any more, so remove it.
+                        new_lc_map.remove(&lc_index);
```
Pratyush: Hmm doesn't this delete LCs when they're made into a new variable? For inlining it doesn't matter, because they're substituted directly into the actual constraints. But here, if the LC is "outlined", we cannot delete it from `new_lc_map`, no?

weikengchen: Actually, we can. 

If an LC is "outlined", then its entry in `new_lc_map` is now `(1, the new witness variable)`. 

When it is not used anymore (aka, triggering this code), it means that all the subsequent LCs that use this LC has inlined this LC, by extending a term `(1 * coeff, the new witness variable)` into themselves. So, it can be safely deleted from the `new_lc_map`.

And this does not affect the code at about line 411, which creates additional constraints. This is because we copy the "inlined" LC data (not its index) in `new_witness_linear_combinations`.

Pratyush: Ah I misunderstood; this works because the witness *variable* exists, and a corresponding constraint, and so the LC is useless. cool!

In that case, we should definitely extract this common logic between `inline_all_lcs` and `outline_lcs` (by taking in `new_lc_maps`), and leave the above explanation as a comment above the use in `outline_lcs`.

weikengchen: Like this? Or more than that?
```
// All the LCs depending on `lc_index` have inlined it.
//
// If `lc_index` is outlined, then its witness variable has 
// replaced the original LC.
//
// This lc is not used any more, so remove it.
```

Pratyush: How about something like this?

```rust 
// Deletion is safe for both outlining and inlining:
// * Inlining: the LC is substituted directly into all use 
//   sites, and so once it is fully inlined, it is redundant.
// * Outlining: the LC is associated with a new variable `w`,
//   and a new constraint of the form `lc_data * 1 = w`, where
//   `lc_data` is the actual data in the linear combination.
//   Furthermore, we replace its entry in `new_lc_map` with `(1, w)`.
//   Once `w` is fully inlined, then we can delete the entry from `new_lc_map`
```

weikengchen: Needs a period at the end of the comments. 

Add this comment to BOTH `inline_all_lcs` and `outline_lcs`, or?

weikengchen: And it makes sense!

Pratyush: Let's extract the common logic into a method, and then call this method from `inline_all_lcs` and `outline_lcs`?

weikengchen: Let me have a try on the common logic. Very likely, `inline_all_lcs` and `outline_lcs` will provide a stateful function to this common interface. The common interface asks this function to rewrite each LC, and this function can keep a state.

Pratyush: Yup!

relations/src/r1cs/constraint_system.rs
```diff
@@ -309,91 +310,111 @@ impl<F: Field> ConstraintSystem<F> {
...
+        // the early ones, and decides whether or not to dedicate a witness
+        // variable for this LC.
+        //
+        // If true, the LC is replaced with 1 * this witness variable.
+        // Otherwise, the LC is inlined.
+        //
+        let mut new_lc_map = BTreeMap::new();
         for (&index, lc) in self.lc_map.iter() {
-            let mut outlined_lc = LinearCombination::new();
+            // Inline the LC.
```
ValarDragon:
```suggestion
            // Inline the LC, unwrapping symbolic LCs that may constitute it,
            // and updating them according to outlinings in prior iterations.
```

relations/src/r1cs/constraint_system.rs
```diff
@@ -309,91 +310,111 @@ impl<F: Field> ConstraintSystem<F> {
...
+
+        // `new_lc_map` stores the modified linear combinations.
+        //
+        // This loop goes through all the LCs in the map, starting from
+        // the early ones, and decides whether or not to dedicate a witness
+        // variable for this LC.
+        //
+        // If true, the LC is replaced with 1 * this witness variable.
+        // Otherwise, the LC is inlined.
+        //
```
ValarDragon:
```suggestion
        // `new_lc_map` stores the modified linear combinations.
        //
        // This loop goes through all the LCs in the map, starting from
        // the early ones, and decides whether or not to dedicate a witness
        // variable for this LC.
        //
        // If true, the LC is replaced with 1 * this witness variable.
        // Otherwise, the LC is inlined.
        //
        // Each iteration first updates the LC according to outlinings in prior iterations,
        // and then sees if it should be outlined, and if so adds the outlining to the map.
        //
```

relations/src/r1cs/constraint_system.rs
ValarDragon: This doesn't matter for soundness right? Its just for efficiency?

ValarDragon (approved):
LGTM! Left a couple updates to the comments that helped me understand the structure of the loop. Thanks for fixing the bugs, and significantly improving the documentation around this 

In a future PR, I think we should be able to improve the efficiency by changing `new_lc_map` to being just a map of the differences (`SymbolicLC -> Var`). (Instead of making a copy of all the LC's)

relations/src/r1cs/constraint_system.rs
Pratyush: Yes, for memory consumption.

weikengchen:
Thanks for the review. I will resolve the comments soon and also merge the logic of the two functions for fewer repetitions.

(I might have some delays due to sending the Mac for fixing---my "e" key is now an "ee" key.)

weikengchen:
Response to this:

> In a future PR, I think we should be able to improve the efficiency by changing new_lc_map to being just a map of the differences (SymbolicLC -> Var). (Instead of making a copy of all the LC's)

Indeed almost all lcs would be rewritten due to inlining (all symbolic LCs are replaced by linear combinations of input/witness variables).

weikengchen:
The implementation has been refactored in that the common elements between `inline_all_lcs` and `outline_lcs` are extracted.

@Pratyush @ValarDragon You may want to give a fresh reading to see if the logic flows right.

Note: The PR is not ready for merging, as I need to test it on Groth16 and Marlin.

weikengchen:
And also, should we change `outline_lcs` to `outline_all_lcs`?

weikengchen:
A bug has been fixed. It now passes on Groth16 and Marlin.

Pratyush (approved):
This PR looks great now, thanks for the effort =)

weikengchen:
Opinions on changing `outline_lcs` to `outline_all_lcs`? This is a good chance to decide (since it has not been used anywhere else).

Pratyush:
I think `outline_lcs` is better, as we're not outlining all the LCs, only those that are large, no?

weikengchen:
That makes sense!

ValarDragon:
This function is presumably not an external facing API right?

The external facing api should probably be "reduce_matrix_density"

weikengchen:
Indeed it will be externally facing, in that Marlin calls outline_lcs instead of inline_all_lcs before converting the constraint system into a matrix.

Technically, outline_lcs is also mostly inlining.

ValarDragon:
I see. I think we should make this function private, and make the external facing function `reduce_matrix_arity` (or `reduce_constraint_density` or something), which will then call this function. In the future, other matrix arity optimizations may make it in there.

ValarDragon (approved):
Looks great to me

weikengchen:
I would call it `reduce_constraint_weight` since Ale uses `weight` in our paper on recursive Marlin.

weikengchen:
Ok. The final change passes the CI.

I hereby need one person with write access to merge this PR.

Pratyush:
In that case, should we provide a higher level API to inline_all_lcs too?

perhaps a single “finalize” method that takes in a rename OptimizationMode enum and then internally calls inline_lcs or outline_lcs? (This can be a separate PR)

PR #334, opened 2020-12-31
Finalize

### LINKED ISSUE #333: provide a higher level API for CS finalization
should we provide a single “finalize” method that takes in a rename OptimizationMode enum and then internally calls inline_lcs or outline_lcs? (This can be a separate PR)

_Originally posted by @Pratyush in https://github.com/arkworks-rs/snark/issues/331#issuecomment-751516304_

npwardberkeley:
Initial attempt to resolve #333

### DIFF
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -148,7 +148,7 @@ impl<F: Field> ConstraintSystem<F> {
                 construct_matrices: true,
             },
 
-            optimization_goal: OptimizationGoal::None,
+            optimization_goal: OptimizationGoal::Constraints,
         }
     }
 
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -47,6 +47,10 @@ pub struct ConstraintSystem<F: Field> {
     /// The number of linear combinations
     pub num_linear_combinations: usize,
 
+    /// The parameter we aim to minimize in this constraint system (either the
+    /// number of constraints or their total weight).
+    pub optimization_goal: OptimizationGoal,
+
     /// Assignments to the public input variables. This is empty if `self.mode
     /// == SynthesisMode::Setup`.
     pub instance_assignment: Vec<F>,
@@ -91,6 +95,16 @@ pub enum SynthesisMode {
     },
 }
 
+/// Defines the parameter to optimize for a `ConstraintSystem`.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub enum OptimizationGoal {
+    /// Minimize the number of constraints.
+    Constraints,
+    /// Minimize the total weight of the constraints (the number of nonzero
+    /// entries across all constraints).
+    Weight,
+}
+
 impl<F: Field> ConstraintSystem<F> {
     #[inline]
     fn make_row(&self, l: &LinearCombination<F>) -> Vec<(F, usize)> {
@@ -109,7 +123,7 @@ impl<F: Field> ConstraintSystem<F> {
             .collect()
     }
 
-    /// Construct an ampty `ConstraintSystem`.
+    /// Construct an empty `ConstraintSystem`.
     pub fn new() -> Self {
         Self {
             num_instance_variables: 1,
@@ -131,6 +145,8 @@ impl<F: Field> ConstraintSystem<F> {
             mode: SynthesisMode::Prove {
                 construct_matrices: true,
             },
+
+            optimization_goal: OptimizationGoal::Constraints,
         }
     }
 
@@ -147,6 +163,12 @@ impl<F: Field> ConstraintSystem<F> {
     /// Check whether `self.mode == SynthesisMode::Setup`.
     pub fn is_in_setup_mode(&self) -> bool {
         self.mode == SynthesisMode::Setup
+    }
+
+    /// Check whether this constraint system aims to optimize weight,
+    /// rather than number of constraints.
+    pub fn is_optimizing_weight(&self) -> bool {
+        self.optimization_goal == OptimizationGoal::Weight
     }
 
     /// Check whether or not `self` will construct matrices.
@@ -481,8 +503,17 @@ impl<F: Field> ConstraintSystem<F> {
     /// Useful for SNARKs like [\[Marlin\]](https://eprint.iacr.org/2019/1047) or
     /// [\[Fractal\]](https://eprint.iacr.org/2019/1076), where addition gates
     /// are not cheap.
-    pub fn reduce_constraint_weight(&mut self) {
+    fn reduce_constraint_weight(&mut self) {
         self.outline_lcs();
+    }
+
+    /// Finalize the constraint system (either by outlining or inlining, depending
+    /// on the set optimization mode).
+    pub fn finalize(&mut self) {
+        match self.optimization_goal {
+            OptimizationGoal::Constraints => self.inline_all_lcs(),
+            OptimizationGoal::Weight => self.outline_lcs(),
+        };
     }
 
     /// This step must be called after constraint generation has completed, and
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -176,6 +176,12 @@ impl<F: Field> ConstraintSystem<F> {
     /// Specify whether this constraint system should aim to optimize weight,
     /// number of constraints, or neither.
     pub fn set_optimization_goal(&mut self, goal: OptimizationGoal) {
+        // `set_optimization_goal` should only be executed before any constraint or value is created.
+        assert_eq!(self.num_instance_variables, 1);
+        assert_eq!(self.num_witness_variables, 0);
+        assert_eq!(self.num_constraints, 0);
+        assert_eq!(self.num_linear_combinations, 0);
+
         self.optimization_goal = goal;
     }
 
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -266,7 +266,8 @@ impl<F: Field> ConstraintSystem<F> {
         Ok(())
     }
 
-    /// Count the number of times each LC is used within other LCs in the constraint system
+    /// Count the number of times each LC is used within other LCs in the
+    /// constraint system
     fn lc_num_times_used(&self, count_sinks: bool) -> Vec<usize> {
         let mut num_times_used = vec![0; self.lc_map.len()];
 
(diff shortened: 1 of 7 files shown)


weikengchen:
This fmt issue may have something to do with the Rust version, aka
- is it 1.48.0?
- is it stable?

Because some other versions, and some nightly, have different cargo fmt preferences.

weikengchen:
let me cc @Pratyush and @ValarDragon here for some feedback on the API.

relations/src/r1cs/constraint_system.rs
```diff
@@ -149,6 +165,18 @@ impl<F: Field> ConstraintSystem<F> {
         self.mode == SynthesisMode::Setup
     }
 
+    /// Check whether this constraint system aims to optimize weight or
+    /// number of constraints.
+    pub fn get_optimization_goal(&self) -> OptimizationGoal {
```
Pratyush:
```suggestion
    pub fn optimization_goal(&self) -> OptimizationGoal {
```

relations/src/r1cs/constraint_system.rs
```diff
@@ -481,10 +512,19 @@ impl<F: Field> ConstraintSystem<F> {
     /// Useful for SNARKs like [\[Marlin\]](https://eprint.iacr.org/2019/1047) or
     /// [\[Fractal\]](https://eprint.iacr.org/2019/1076), where addition gates
     /// are not cheap.
-    pub fn reduce_constraint_weight(&mut self) {
+    fn reduce_constraint_weight(&mut self) {
         self.outline_lcs();
     }
```
Pratyush: We can remove this now, right?

relations/src/r1cs/constraint_system.rs
```diff
@@ -782,6 +822,23 @@ impl<F: Field> ConstraintSystemRef<F> {
...
 
+    /// Check whether this constraint system aims to optimize weight or
+    /// number of constraints.
+    #[inline]
+    pub fn get_optimization_goal(&self) -> OptimizationGoal {
+        self.inner().map_or(OptimizationGoal::Constraints, |cs| {
+            cs.borrow().get_optimization_goal()
+        })
+    }
```
Pratyush:
```suggestion
    pub fn optimization_goal(&self) -> OptimizationGoal {
        self.inner().map_or(OptimizationGoal::Constraints, |cs| {
            cs.borrow().optimization_goal()
        })
    }
```

relations/src/r1cs/constraint_system.rs
```diff
@@ -91,6 +95,16 @@ pub enum SynthesisMode {
     },
 }
 
+/// Defines the parameter to optimize for a `ConstraintSystem`.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub enum OptimizationGoal {
```
ValarDragon, line 100: Should we have a `None` here, which will do no optimization passes?

relations/src/r1cs/constraint_system.rs
```diff
@@ -260,16 +289,17 @@ impl<F: Field> ConstraintSystem<F> {
...
-    ///     (&ConstraintSystem<F>, usize, &mut LinearCombination<F>)
+    /// This method is used as a subroutine of `inline_all_lcs` and
+    /// `outline_lcs`.
     ///
-    /// The transformer function returns the number of new witness variables needed
-    /// and a vector of new witness assignments (if not in the setup mode).
-    ///     (usize, Option<Vec<F>>)
+    /// The transformer function is given a references of this constraint system
+    /// (&self), number of times used, and a mutable reference of the linear
+    /// combination to be transformed.     (&ConstraintSystem<F>, usize,
```
ValarDragon: Were these comments re-ordered by a linter? Its kinda weird imo that the tuple is split across a new line

relations/src/r1cs/constraint_system.rs
npwardberkeley: Wasn't the idea to add other weight-reducing optimizations, apart from just outlining? (That's what the comment here says.) In which case we'd still want this as a separate function

relations/src/r1cs/constraint_system.rs
npwardberkeley: That makes sense. Do you think `None`, or `Constraints`, should be the default?

ValarDragon: I think constraints should be the default

weikengchen:
It seems that all the comments have been resolved, with this one needing more discussion:

whether or not to keep this function
```
fn reduce_constraint_weight(&mut self) {
```

Since this function is now `fn` rather than `pub fn`, I think for now it is okay to remove this function. The function is only, previously used by Marlin (who needs to change accordingly).

Nick, what do you think?

npwardberkeley:
Yes, that makes sense!

weikengchen:
Let me cc @Pratyush and @ValarDragon for a final pass

relations/src/r1cs/constraint_system.rs
Pratyush: What are the semantics of calling finalize with None? In the end we have to get rid of all SymbolicLC before converting to matrices

Pratyush:
I left a comment about None

relations/src/r1cs/constraint_system.rs
weikengchen: I think `None` would still need to inline Symbolic LC (`self.inline_all_lcs()`, without which `to_matrices` would fail.

Dev, what do you think? Basically, at this moment, `None` would be the same as `Constraints`, and it is uncertain if in the future we will treat them differently.

ValarDragon: Right, but we wouldn't do any additional passes for inlining or outlining. Not sure if theres a real use case for that though. (Maybe it depends on if in-lining / out-lining is expensive relative to circuit generation in any context?)

weikengchen: I think given that no specific use cases pop up for `None`, and we currently do not have a good idea of faster inlining (without optimization to reduce constraints), maybe we can omit `None` and add it later. Dev, what do you think?

ValarDragon: Sounds good to me!

relations/src/r1cs/constraint_system.rs
```diff
@@ -149,6 +167,18 @@ impl<F: Field> ConstraintSystem<F> {
...
+    /// Check whether this constraint system aims to optimize weight,
+    /// number of constraints, or neither.
+    pub fn optimization_goal(&self) -> OptimizationGoal {
+        self.optimization_goal
+    }
+
+    /// Specify whether this constraint system should aim to optimize weight,
+    /// number of constraints, or neither.
+    pub fn set_optimization_goal(&mut self, goal: OptimizationGoal) {
+        self.optimization_goal = goal;
```
ValarDragon, line 185: This API makes it seem to me that a circuit can flip-flop to having different optimization goals at different points in circuit generation. Is this intentional?

If not, perhaps the goal should be added at CS initialization, and only have a getter?

weikengchen: You are right. Actually this needs to be fixed for the circuit, i.e., can be set only once, before any constraints or variables are generated.

My proposal is to enforce that `set_optimization_goal` would panic if there is any constraint or variable in the constraint system.

weikengchen: Putting it as the initialization is a little hard,  for two reasons: (1) we would better not to change the process of creating a cs since it would be overly breaking, yet Rust does not allow a default/optional argument; (2) there is more than one way to create a cs (e.g. by ref).

Pratyush: In the long term my idea was to make ConstraintSystem private somehow, and only have ConstraintSystemRef. Then, on ConstraintSystemRef we can have a new constructor “with_optimization_goal” which takes an input a OptimizationGoal and outputs a ConstraintSystemRef with that goal

relations/src/r1cs/constraint_system.rs
ValarDragon: Actually, I do think `None` is still useful, in that it could just make far fewer symbolic LC's for memory reduction. E.g. when multiplying by a constant, we don't need to make a new symbolic LC if we don't care about optimizations

ValarDragon: And we could have drop semantics to symbolic LC's to remove intermediate variables from the map as soon as they're dropped from memory. It is a significant engineering effort though, so maybe not worth having it now. (Its non breaking to add it later)

weikengchen: That makes sense. Then let us keep `None` and, for now, let `None` also do `inline_all_lcs`. More changes may be coming in the future.

relations/src/r1cs/constraint_system.rs
weikengchen: That is a good direction. Then, I think, for this PR, we can keep the current `set_optimization_goal` function for now (which I assume will only be used in Marlin). In the future, Marlin would need to call that new constructor, and setting an optimization goal would only be permitted during the initialization.

Making ConstraintSystem private seems to be a larger and breaking change.

relations/src/r1cs/constraint_system.rs
```diff
@@ -473,16 +504,14 @@ impl<F: Field> ConstraintSystem<F> {
...
-    /// Useful for SNARKs like [\[Marlin\]](https://eprint.iacr.org/2019/1047) or
-    /// [\[Fractal\]](https://eprint.iacr.org/2019/1076), where addition gates
-    /// are not cheap.
-    pub fn reduce_constraint_weight(&mut self) {
-        self.outline_lcs();
+    /// Finalize the constraint system (either by outlining or inlining,
+    /// if an optimization goal is set).
+    pub fn finalize(&mut self) {
+        match self.optimization_goal {
+            OptimizationGoal::None => (),
```
weikengchen: This line would need to be changed into `OptimizationGoal::None =>  self.inline_all_lcs(),`

weikengchen (changes requested):

weikengchen:
Based on the discussion above, I think we will keep `None` for now and redirect it to `self.inline_all_lcs()`. This is the last change needed/ In the future, different treatments may occur, such as reducing the amount of symbolic LC.

weikengchen:
Checks pass :) Pratyush and Dev, if it looks good, feel free to merge or do some final edits.

relations/src/r1cs/constraint_system.rs
ValarDragon: This sounds good to me, but can we add a check to ensure that it doesn't change mid circuit generation?

weikengchen:
One more: we need to export OptimizationGoal here:

https://github.com/npwardberkeley/snark/blob/finalize/relations/src/r1cs/mod.rs#L21

Otherwise, Marlin could not find it.

relations/src/r1cs/constraint_system.rs
weikengchen: Like enforcing (assert) the following?

            num_instance_variables: 1,
            num_witness_variables: 0,
            num_constraints: 0,
            num_linear_combinations: 0,

weikengchen:
add one to enforce that `set_optimization_goal` must check that no constraint/variable has been written.

```
        num_instance_variables: 1,
        num_witness_variables: 0,
        num_constraints: 0,
        num_linear_combinations: 0,
```

(sorry to bother Nick---I could add them if you add me as a PR repo's writer)

relations/src/r1cs/constraint_system.rs
ValarDragon: Yeah!

weikengchen (approved):
The final changes have been applied. Going to merge and updating Marlin then.

Issue #336, opened 2021-01-03
Small Vec Linear Combinations


ValarDragon:
## Summary

As currently written, a symbolic LC is created on every operation. We should expect that most Symbolic LC's are small, and probably should try to keep things on stack via small vec optimizations, perhaps using https://github.com/servo/rust-smallvec.

We definitely shouldn't switch without a more detailed benchmark suite (getting average symbolic LC size, seeing how much of the time is spent in mallocs, seeing if things actually get faster etc.)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

weikengchen:
A separate proposal would be to avoid the generation of too many symbolic LC.

Currently, each AllocatedFp is associated with a Variable in the constraint system. Thus, mul_by_constant and add would all create a new symbolic LC.

One potential direction is to let AllocatedFp uses a linear combination of variables instead of a symbolic LC, if it is optimized for the number of constraints.

ValarDragon:
> One potential direction is to let AllocatedFp uses a linear combination of variables instead of a symbolic LC, if it is optimized for the number of constraints.

I think this is a great idea. The alternative idea I was thinking of was to introduce drop semantics to variables, but that was growing in complexity very rapidly.  (We may still need something of that form to remove symbolic LC's that are intermediate computation steps though, in the case of density optimizations)

Pratyush:
> One potential direction is to let AllocatedFp uses a linear combination of variables instead of a symbolic LC, if it is optimized for the number of constraints.

One way to do this is to change the `Variable` enum as follows:
```rust
pub enum Variable<F> {
	/// Represents the "zero" constant.
    Zero,
    /// Represents of the "one" constant.
    One,
    /// Represents a public instance variable.
    Instance(usize),
    /// Represents a private witness variable.
    Witness(usize),
    /// Represents of a linear combination.
    SymbolicLc(InlineOrIndex<F>),
}

// Needed if we want to make this an implementation detail.
pub struct InlineOrIndex<F>(InlineOrIndexInner<F>);

enum InlineOrIndexInner<F> {
	Index(usize),
	Lc(StackVec<(F, Variable<F>)>), // Not sure if this needs to be `Box`
}
```

[...]

PR #339, opened 2021-01-05
Add helper method to inline `mul_by_const` immediately under constraints optimization goal

ValarDragon:
## Description

This adds a method to constraint system that gadgets such as FpVar would call to do multiplication by a constant. It handles inlining immediately when under the `Constraints` Optimization goal, and doesn't create additional symbolic variables when the variable is `Zero`. This could cause speed losses to constraint-optimized circuits when doing two mul by consts in a row, but I think such a case isn't that likely.

This is intended to be used here: https://github.com/arkworks-rs/r1cs-std/blob/master/src/fields/fp/mod.rs#L192

---

write a little note why.

- [ ] Wrote unit tests - its hard to write unit tests in this repo
- [x] Updated relevant documentation in the code
- [x] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Improvements
 - #325 Reduce memory consumption during inlining
+- #339 Reduce memory consumption under `Constraints` optimization goal
 
 ### Bug fixes
 
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -256,8 +256,8 @@ impl<F: Field> ConstraintSystem<F> {
         var: &Variable,
         other: &F,
     ) -> crate::r1cs::Result<Variable> {
-        // Create a new symbolic LC using this constant, deffering the inlining
-        // until the finalize phase
+        // Create a new symbolic LC using this constant, deferring the inlining
+        // until the finalization phase.
         #[inline]
         fn base_mul_by_constant<F: Field>(
             cs: &mut ConstraintSystem<F>,
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -256,7 +256,8 @@ impl<F: Field> ConstraintSystem<F> {
         var: &Variable,
         other: &F,
     ) -> crate::r1cs::Result<Variable> {
-        // Create a new symbolic LC using this constant
+        // Create a new symbolic LC using this constant, deffering the inlining
+        // until the finalize phase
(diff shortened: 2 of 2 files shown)


weikengchen:
Corrected one small typo `deffering` -> `deferring`. (note: if you are using a Mac, you might be a victim of the bufferfly keyboard like me  https://support.apple.com/keyboard-service-program-for-mac-notebooks)

Others look good.

weikengchen (approved):

Pratyush:
Do we need to make a new API for this? As in, inside `new_lc`, we can detect if the incoming linear combination consists of just a single term, and can inline it directly if we're in the `Constraints` optimization goal.

(This might even be helpful under *any* optimization goal, when the variable in the single term is not `SymbolicLc`).

ValarDragon:
That works, though it is unfortunate that we'd be doing wasted heap allocations then, versus producing the right lc. (One extra heap allocation for basically every operation) We don't really have empirical verification that thats a bottleneck though.

Pratyush:
If we make `LinearCombination` to be a `SmallVec` of size 1 then we don't have to pay that overhead at all =)

ValarDragon:
True! I'll close this in favor of going with that approach

ValarDragon:
Actually, I'm realizing now that this actually increases memory overhead in the current design, since we're still storing each intermediate LC, and now those LC's are longer. (as would the `new_lc` approach) This type of approach would be useful if there were a drop-semantics style approach of removing old lc's from the lc_map.

Pratyush:
I think there are two approaches to handle this:
* Either we have linear combinations in the variable directly, or in the `lc_map`, but not in both. This way, if we inline LCs, then when the variable goes out of scope, the memory for the LC is freed immediately.

* We hold a `Rc<RefCell<LcMap>>` or something inside a `SymbolicLc`, and when the `SymbolicLc` gets dropped, we look inside the reference to `LcMap` and do some garbage collection there. Not sure how expensive this would be, though.

ValarDragon:
We had the same ideas =) #cref https://github.com/arkworks-rs/snark/issues/336#issuecomment-754852431

> Either we have linear combinations in the variable directly, or in the lc_map, but not in both. This way, if we inline LCs, then when the variable goes out of scope, the memory for the LC is freed immediately.

We still need LC's getting saved to the lc_map at some point though, in order to do density saving optimizations. Doing that at constraint generation causes some awkward mutability problems though.

Pratyush:
Let's consolidate in #336

PR #347, opened 2021-04-04
Add `into_inner` function for `ConstraintSystemRef<F>`

ryanleh:
## Description

Adds a helper function to unwrap a `ConstraintSystemRef<F>`. This is useful for managing memory in https://github.com/arkworks-rs/marlin/pull/71.

---

write a little note why.

- [ ] Wrote unit tests
- [x] Updated relevant documentation in the code
- [x] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`

### DIFF
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -772,7 +772,7 @@ impl<F: Field> ConstraintSystemRef<F> {
     pub fn into_inner(self) -> ConstraintSystem<F> {
         match self {
             Self::CS(a) => Rc::try_unwrap(a).unwrap().into_inner(),
-            Self::None => panic!()
+            Self::None => panic!(),
         }
     }
 
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -767,12 +767,13 @@ impl<F: Field> ConstraintSystemRef<F> {
         }
     }
 
-    /// Consumes self to return the inner `ConstraintSystem<F>`. Panics if
-    /// any other references to `self` exist.
-    pub fn into_inner(self) -> ConstraintSystem<F> {
+    /// Consumes self to return the inner `ConstraintSystem<F>`. Returns 
+    /// `None` if `Self::CS` is `None` or if any other references to
+    /// `Self::CS` exist.  
+    pub fn into_inner(self) -> Option<ConstraintSystem<F>> {
         match self {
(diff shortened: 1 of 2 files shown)


weikengchen:
Question: the comment said that it "panics if any other references to `self` exist."

From the code, it panics in the case when `ConstraintSystemRef` does not contain a ConstraintSystem?

Pratyush:
Maybe just have the method return an Option? That way the decision to panic is left to users

ValarDragon:
I think going with an option is better as well

relations/src/r1cs/constraint_system.rs
```diff
@@ -767,6 +767,16 @@ impl<F: Field> ConstraintSystemRef<F> {
         }
     }
 
+    /// Consumes self to return the inner `ConstraintSystem<F>`. Returns
+    /// `None` if `Self::CS` is `None` or if any other references to
+    /// `Self::CS` exist.  
+    pub fn into_inner(self) -> Option<ConstraintSystem<F>> {
+        match self {
+            Self::CS(a) => Some(Rc::try_unwrap(a).unwrap().into_inner()),
```
Pratyush: Shouldn’t this just use a `map`?

weikengchen: Updated with map in https://github.com/arkworks-rs/snark/pull/347/commits/1cfae51ff8438f77fd6876e3f7901cd357944609. Have to use a clone on `Rc` but I guess it should be okay.

weikengchen:
Ready for a final pass.

relations/src/r1cs/constraint_system.rs
```diff
@@ -767,6 +767,14 @@ impl<F: Field> ConstraintSystemRef<F> {
         }
     }
 
+    /// Consumes self to return the inner `ConstraintSystem<F>`. Returns
+    /// `None` if `Self::CS` is `None` or if any other references to
+    /// `Self::CS` exist.  
+    pub fn into_inner(self) -> Option<ConstraintSystem<F>> {
+        self.inner()
+            .map(|cs| Rc::try_unwrap(cs.clone()).unwrap().into_inner())
```
Pratyush: Sorry I meant something like this, the goal being to avoid panics if more than one reference to the inner `cs` exists.

```suggestion
		match self {
			Self::CS(a) => Rc::try_unwrap(a).ok().map(|s| s.into_inner()),
			Self::None => None,
		}
```

PR #348, opened 2021-04-20
Add `Relation` trait

### LINKED ISSUE #321: Remove dependence of `SNARK` on `ConstraintSynthesizer`
`ConstraintSynthesizer` is specific to `R1CS`, so either we need to generalize `ConstraintSynthesizer` to other relations that might pop up in the future, or we need to remove it.

### LINKED ISSUE #323: Make `ConstraintSynthesizer::generate_constraints` not consume `self` by value

### LINKED ISSUE #341: Generate public inputs for verifier also via ConstraintSystem API
## Problem Definition

Right now, when the verifier needs to verify the public input, they have to implement and call `ToConstraintField` on the input. This is error prone for two reasons:
* the order of calling `ToConstraintField` on various parts of the public input must match exactly the order of `new_input` calls in the `ConstraintSynthesizer`. 
* the impl inside `ToConstraintField` must match that of `new_input` exactly, leading to another source of errors.

## Proposal

I propose that we instead leverage the existing `ConstraintSystem` API to generate these inputs: we add a `Verifier` variant to the `Mode` enum, which generates variable assignments *only* for instance/input variables.

My only worry is that this could cause extra memory/computation for the verifier, but we can at least try it and see. I have two ideas to minimize this overhead:
* a way to minimize these costs alternative would be to provide `ConstraintSynthesizer` with a `finalize_public_input_generation` method that allows it to specify when it has completed generating public inputs, so that it can early return.

* Make a related instantiation of `ConstraintSynthesizer` that only generates public inputs, and does not have the rest of the circuit logic. This is still better than the `ToConstraintField` idea because it is easy to match up the `new_input` statements on both sides.

____

#### For Admin Use

- [X] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

### LINKED ISSUE #342: Low-level API for SNARK trait
## Summary

Introduce a low-level API for setup, indexing, proving, and verifying that directly reasons about the relation, instead of going via our `ConstraintSystem` API.

## Problem Definition

Right now, for proving R1CS via our `SNARK` traits, we have to go via the `ConstraintSynthesizer` (and hence `ConstraintSystem`) trait. This is unsatisfactory for a couple of reasons:

1) Using our libraries with external R1CS formats like `zkinterface` incurs performance overheads because we have to convert to `ConstraintSystem` and then back to matrices, instead of directly reading the matrices from the external format.
2) The `relations` crate is at the moment more about data structures for working with a particular relation (R1CS) rather than about the relation itself. For example, the `R1CS` relation consists of `(i, x, w)` where `i` consists of the `R1CS` matrices, and `x` and `w` are the public input and witness, respectively. However, the current `ark_relations::r1cs` module doesn't have any data structure reflecting these, and only has data structures like `ConstraintSystemRef`.

## Proposal

* Add a `Relation` trait in `relations` that looks like:
```rust
pub trait Relation {
	type Index;
	type Instance;
	type Witness;
	
	fn check_membership(i: &Self::Index, x: &Self::Instance, w: &Self::Witness) -> bool;
}
```
* Modify the `SNARK` trait as follows:
```rust
pub trait SNARK<R: Relation> {
	fn index(pp: &Self::Parameters, i: &R::Index) -> (Self::ProvingKey, Self::VerifyingKey);
	// same for proving and verifying
}
```
Additionally, we add a new R1CS-specific trait:
```rust
pub trait R1CSSnark: SNARK<R1CS> {
	fn index_from_cs<CS: ConstraintSynthesizer>(pp: &Self::Parameters, cs: CS) -> (Self::ProvingKey, Self::VerifyingKey) {
		// default impl using the `SNARK::index`, by `calling cs.into_matrices()`. 
	}
}
```
(We might need equivalents for `PreprocessingSNARK`.)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Pratyush:
## Description

Adds two traits: `Relation` and `NPRelation`. Modifies the `SNARK` trait to use these, instead of the `ConstraintSystem` infrastructure. 

TBD: How to integrate the old `ConstraintSystem`-based APIs into this. See https://github.com/arkworks-rs/snark/pull/348#discussion_r616286265 for an idea.

closes #323
closes #321
closes #341
closes #342

---

write a little note why.

- [X] Targeted PR against correct branch (master)
- [X] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests --- N/A
- [X] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [X] Re-reviewed `Files changed` in the Github PR explorer

### DIFF
--- a/relations/src/lib.rs
+++ b/relations/src/lib.rs
@@ -16,3 +16,21 @@
 extern crate ark_std;
 
 pub mod r1cs;
+
+/// An *indexed relation* is a set of triples of the form `(index, instance, witness)`.
+pub trait Relation {
+    /// The index is a "large" but static part of the triple. Examples include 
+    /// the circuit in the circuit satisfiability relation, and constraint 
+    /// matrices in the R1CS relation.
+    type Index: Eq;
+    /// The instance is a "small" part of the triple. Like the index, it is publicly known.
+    type Instance: Eq;
+    /// The instance is a "large" but private part of the triple.
+    type Witness: Eq;
+}
+
+/// An *indexed NP relation* is a relation with an efficient membership check.
+pub trait NPRelation: Relation {
+    /// Checks whether the triple `(index, instance, witness)` is a member of the relation.
+    fn check_membership(index: &Self::Index, instance: &Self::Instance, witness: &Self::Witness) -> bool;
+}
--- a/relations/src/lib.rs
+++ b/relations/src/lib.rs
@@ -25,7 +25,7 @@ pub trait Relation {
     type Index: Eq;
     /// The instance is a "small" part of the triple. Like the index, it is publicly known.
     type Instance: Eq;
-    /// The instance is a "large" but private part of the triple.
+    /// The witness is a "large" but private part of the triple.
     type Witness: Eq;
 }
 
--- a/relations/src/lib.rs
+++ b/relations/src/lib.rs
@@ -19,8 +19,8 @@ pub mod r1cs;
 
 /// An *indexed relation* is a set of triples of the form `(index, instance, witness)`.
 pub trait Relation {
-    /// The index is a "large" but static part of the triple. Examples include 
-    /// the circuit in the circuit satisfiability relation, and constraint 
+    /// The index is a "large" but static part of the triple. Examples include
+    /// the circuit in the circuit satisfiability relation, and constraint
     /// matrices in the R1CS relation.
     type Index: Eq;
     /// The instance is a "small" part of the triple. Like the index, it is publicly known.
@@ -32,5 +32,9 @@ pub trait Relation {
 /// An *indexed NP relation* is a relation with an efficient membership check.
 pub trait NPRelation: Relation {
     /// Checks whether the triple `(index, instance, witness)` is a member of the relation.
-    fn check_membership(index: &Self::Index, instance: &Self::Instance, witness: &Self::Witness) -> bool;
+    fn check_membership(
+        index: &Self::Index,
+        instance: &Self::Instance,
+        witness: &Self::Witness,
+    ) -> bool;
 }
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -1,6 +1,6 @@
 #[cfg(feature = "std")]
 use crate::r1cs::ConstraintTrace;
-use crate::r1cs::{LcIndex, LinearCombination, Matrix, SynthesisError, Variable};
+use crate::r1cs::{LcIndex, LinearCombination, SynthesisError, Variable, ConstraintMatrices};
 use ark_ff::Field;
 use ark_std::{
     any::{Any, TypeId},
@@ -642,36 +642,6 @@ impl<F: Field> ConstraintSystem<F> {
             }
         }
     }
-}
-/// The A, B and C matrices of a Rank-One `ConstraintSystem`.
-/// Also contains metadata on the structure of the constraint system
-/// and the matrices.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct ConstraintMatrices<F: Field> {
-    /// The number of variables that are "public instances" to the constraint
-    /// system.
-    pub num_instance_variables: usize,
-    /// The number of variables that are "private witnesses" to the constraint
-    /// system.
-    pub num_witness_variables: usize,
-    /// The number of constraints in the constraint system.
-    pub num_constraints: usize,
-    /// The number of non_zero entries in the A matrix.
-    pub a_num_non_zero: usize,
-    /// The number of non_zero entries in the B matrix.
-    pub b_num_non_zero: usize,
-    /// The number of non_zero entries in the C matrix.
-    pub c_num_non_zero: usize,
-
-    /// The A constraint matrix. This is empty when
-    /// `self.mode == SynthesisMode::Prove { construct_matrices = false }`.
-    pub a: Matrix<F>,
-    /// The B constraint matrix. This is empty when
-    /// `self.mode == SynthesisMode::Prove { construct_matrices = false }`.
-    pub b: Matrix<F>,
-    /// The C constraint matrix. This is empty when
-    /// `self.mode == SynthesisMode::Prove { construct_matrices = false }`.
-    pub c: Matrix<F>,
 }
 
 /// A shared reference to a constraint system that can be stored in high level
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -510,9 +510,9 @@ impl<F: Field> ConstraintSystem<F> {
         }
     }
 
-    /// Finalize the constraint system (either by outlining or inlining,
+    /// Optimize the constraint system (either by outlining or inlining,
     /// if an optimization goal is set).
-    pub fn finalize(&mut self) {
+    pub fn optimize(&mut self) {
         match self.optimization_goal {
             OptimizationGoal::None => self.inline_all_lcs(),
             OptimizationGoal::Constraints => self.inline_all_lcs(),
@@ -896,11 +896,11 @@ impl<F: Field> ConstraintSystemRef<F> {
         }
     }
 
-    /// Finalize the constraint system (either by outlining or inlining,
+    /// Optimize the constraint system (either by outlining or inlining,
     /// if an optimization goal is set).
-    pub fn finalize(&self) {
+    pub fn optimize(&self) {
         if let Some(cs) = self.inner() {
-            cs.borrow_mut().finalize()
+            cs.borrow_mut().optimize()
         }
     }
 
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -1,6 +1,8 @@
 #[cfg(feature = "std")]
 use crate::r1cs::ConstraintTrace;
-use crate::r1cs::{ConstraintMatrices, LcIndex, LinearCombination, SynthesisError, Variable};
+use crate::r1cs::{
+    ConstraintMatrices, Instance, LcIndex, LinearCombination, SynthesisError, Variable, Witness,
+};
 use ark_ff::Field;
 use ark_std::{
     any::{Any, TypeId},
@@ -922,6 +924,26 @@ impl<F: Field> ConstraintSystemRef<F> {
     #[inline]
     pub fn to_matrices(&self) -> Option<ConstraintMatrices<F>> {
         self.inner().and_then(|cs| cs.borrow().to_matrices())
+    }
+
+    /// This step must be called after constraint generation has completed, and
+    /// after all symbolic LCs have been inlined into the places that they
+    /// are used.
+    #[inline]
+    pub fn instance_assignment(&self) -> Option<Instance<F>> {
+        self.inner()
+            .map(|cs| cs.borrow().instance_assignment.clone())
+            .map(Instance)
+    }
+
+    /// This step must be called after constraint generation has completed, and
+    /// after all symbolic LCs have been inlined into the places that they
+    /// are used.
+    #[inline]
+    pub fn witness_assignment(&self) -> Option<Witness<F>> {
+        self.inner()
+            .map(|cs| cs.borrow().witness_assignment.clone())
+            .map(Witness)
     }
 
     /// If `self` is satisfied, outputs `Ok(true)`.
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -1,6 +1,6 @@
 #[cfg(feature = "std")]
 use crate::r1cs::ConstraintTrace;
-use crate::r1cs::{LcIndex, LinearCombination, SynthesisError, Variable, ConstraintMatrices};
+use crate::r1cs::{ConstraintMatrices, LcIndex, LinearCombination, SynthesisError, Variable};
 use ark_ff::Field;
 use ark_std::{
     any::{Any, TypeId},
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -96,7 +96,7 @@ pub enum SynthesisMode {
     /// Indicate to the `ConstraintSystem` that it should only generate
     /// constraint matrices and not populate the variable assignments.
     Setup,
-    /// Indicate to the `ConstraintSystem` that it should populate the witnessvariable
+    /// Indicate to the `ConstraintSystem` that it should populate the witness variable
     /// assignments. If additionally `construct_matrices == true`, then generate
     /// the matrices as in the `Setup` case.
     Prove {
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -298,7 +298,7 @@ impl<F: Field> ConstraintSystem<F> {
             }
         }
         self.num_constraints += 1;
-        
+
         Ok(())
     }
 
@@ -541,11 +541,15 @@ impl<F: Field> ConstraintSystem<F> {
     /// Optimize the constraint system (either by outlining or inlining,
     /// if an optimization goal is set).
     pub fn optimize(&mut self) {
-        match self.optimization_goal {
-            OptimizationGoal::None => self.inline_all_lcs(),
-            OptimizationGoal::Constraints => self.inline_all_lcs(),
-            OptimizationGoal::Weight => self.outline_lcs(),
-        };
+        // In verify mode we don't have any linear combinations; all variables
+        // are instance variables, and there are no generated constraints
+        if !self.is_in_verify_mode() {
+            match self.optimization_goal {
+                OptimizationGoal::None => self.inline_all_lcs(),
+                OptimizationGoal::Constraints => self.inline_all_lcs(),
+                OptimizationGoal::Weight => self.outline_lcs(),
+            };
+        }
     }
 
     /// This step must be called after constraint generation has completed, and
--- a/relations/src/r1cs/constraint_system.rs
+++ b/relations/src/r1cs/constraint_system.rs
@@ -18,23 +18,18 @@ use ark_std::{
 
 /// Describes how to generate an R1CS index for a given high-level computation.
 pub trait ConstraintGenerator<F: Field> {
-    /// Generates R1CS constraint matrices by modifying [`cs`] in place.
-    /// The result is stored in [`cs`].
-    fn make_constraints(&self, cs: ConstraintSystemRef<F>) -> crate::r1cs::Result<()>;
-}
+    /// Generates R1CS constraint matrices, and, optionally, the variable assignments
+    /// by modifying [`cs`] in place. The result is stored in [`cs`].
+    ///
+    /// Variable assignments are generated if `!cs.is_in_setup_mode()`.
+    fn generate_constraints_and_variable_assignments(
+        &self,
+        cs: ConstraintSystemRef<F>,
+    ) -> crate::r1cs::Result<()>;
 
-/// Describes how to generate an R1CS instance for a given high-level computation.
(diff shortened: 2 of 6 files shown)


relations/src/r1cs/constraint_system.rs
```diff
@@ -20,7 +20,7 @@ use ark_std::{
 // TODO: Think: should we replace this with just a closure?
 pub trait ConstraintSynthesizer<F: Field> {
     /// Drives generation of new constraints inside `cs`.
-    fn generate_constraints(self, cs: ConstraintSystemRef<F>) -> crate::r1cs::Result<()>;
+    fn generate_constraints(&self, cs: ConstraintSystemRef<F>) -> crate::r1cs::Result<()>;
```
Pratyush: Closes #323 

cc @weikengchen

snark/src/lib.rs
```diff
@@ -35,31 +35,25 @@ pub trait SNARK<F: PrimeField> {
...
-
     /// Generates a proof of satisfaction of the arithmetic circuit C (specified
     /// as R1CS constraints).
-    fn prove<C: ConstraintSynthesizer<F>, R: RngCore + CryptoRng>(
+    fn prove<Rng: RngCore + CryptoRng>(
         circuit_pk: &Self::ProvingKey,
-        circuit: C,
-        rng: &mut R,
+        instance: &R::Instance,
+        witness: &R::Witness,
```
Pratyush: Matches the API from the formal definition better.

snark/src/lib.rs
```diff
@@ -35,31 +35,25 @@ pub trait SNARK<F: PrimeField> {
     /// Errors encountered during setup, proving, or verification.
     type Error: 'static + ark_std::error::Error;
 
-    /// Takes in a description of a computation (specified in R1CS constraints),
-    /// and samples proving and verification keys for that circuit.
-    fn circuit_specific_setup<C: ConstraintSynthesizer<F>, R: RngCore + CryptoRng>(
-        circuit: C,
-        rng: &mut R,
-    ) -> Result<(Self::ProvingKey, Self::VerifyingKey), Self::Error>;
-
```
Pratyush, line 44: I moved this algorithm to `CircuitSpecificSetupSNARK`; it doesn't really belong here.

snark/src/lib.rs
```diff
@@ -72,26 +66,23 @@ pub trait SNARK<F: PrimeField> {
...
+    fn setup<Rng: RngCore + CryptoRng>(
+        index: &R::Index,
+        rng: &mut Rng,
+    ) -> Result<(Self::ProvingKey, Self::VerifyingKey), Self::Error>;
 }
 
 /// A helper type for universal-setup SNARKs, which must infer their computation
 /// size bounds.
-pub enum UniversalSetupIndexError<Bound, E> {
+pub enum IndexingError<Bound, E> {
```
Pratyush, line 87: Indexing is an operation that happens only for `UniversalSetupSNARK`s, so we don't need to specify that separately here.

snark/src/lib.rs
```diff
@@ -101,30 +92,56 @@ pub enum UniversalSetupIndexError<Bound, E> {
 
 /// A SNARK with universal setup. That is, a SNARK where the trusted setup is
 /// circuit-independent.
-pub trait UniversalSetupSNARK<F: PrimeField>: SNARK<F> {
+pub trait UniversalSetupSNARK<R: NPRelation>: SNARK<R> {
     /// Specifies how to bound the size of public parameters required to
     /// generate the index proving and verification keys for a given
     /// circuit.
-    type ComputationBound: Clone + Default + Debug;
+    type IndexBound: Clone + Default + Debug;
```
Pratyush, line 105: Slightly more descriptive name.

snark/src/lib.rs
```diff
@@ -101,30 +92,56 @@ pub enum UniversalSetupIndexError<Bound, E> {
...
     /// Specifies the type of universal public parameters.
     type PublicParameters: Clone + Debug;
 
+    /// Specifies the bound size that is necessary and sufficient to 
+    /// generate public parameters for `index`.
+    fn bound_for_index(index: &R::Index) -> Self::IndexBound;
```
Pratyush, line 111: New API that enables computing the bound for a given index.

snark/src/lib.rs
```diff
@@ -101,30 +92,56 @@ pub enum UniversalSetupIndexError<Bound, E> {
...
 
     /// Indexes the public parameters according to the circuit `circuit`, and
     /// outputs circuit-specific proving and verification keys.
-    fn index<C: ConstraintSynthesizer<F>, R: RngCore + CryptoRng>(
+    ///
+    /// This is a *deterministic* method.
+    fn index(
         pp: &Self::PublicParameters,
-        circuit: C,
-        rng: &mut R,
```
Pratyush, line 125: Indexing is now deterministic, as per the formal definition.

snark/src/lib.rs
```diff
@@ -101,30 +92,56 @@ pub enum UniversalSetupIndexError<Bound, E> {
...
         (Self::ProvingKey, Self::VerifyingKey),
-        UniversalSetupIndexError<Self::ComputationBound, Self::Error>,
+        IndexingError<Self::IndexBound, Self::Error>,
     >;
 }
+
+impl<R: NPRelation, S> CircuitSpecificSetupSNARK<R> for S
+where
+    S: UniversalSetupSNARK<R> 
+{
```
Pratyush: Every `UniversalSetupSNARK` is also a `CircuitSpecificSetupSNARK`, so I added this blanket impl here.

snark/src/r1cs.rs
```diff
@@ -0,0 +1,70 @@
+use ark_relations::r1cs::{R1CS, ConstraintSynthesizer};
```
Pratyush: These are specialized APIs for working with the existing `ConstraintSystem`-based API

snark/src/r1cs.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn process_vk(
+        circuit_vk: &Self::VerifyingKey,
+    ) -> Result<Self::ProcessedVerifyingKey, Self::Error>;
+
+    /// Checks that `proof` is a valid proof of the satisfaction of circuit
+    /// encoded in `circuit_pvk`, with respect to the public input `public_input`,
+    /// specified as R1CS constraints.
+    fn verify_with_processed_vk<C: ConstraintSynthesizer<F>>(
+        circuit_pvk: &Self::ProcessedVerifyingKey,
+        public_input_generator: C,
```
Pratyush: The verifier now takes in a `ConstraintSynthesizer` that outputs the assignment to instance variables.

relations/src/r1cs/constraint_system.rs
```diff
@@ -20,7 +20,7 @@ use ark_std::{
 // TODO: Think: should we replace this with just a closure?
 pub trait ConstraintSynthesizer<F: Field> {
```
Pratyush: Probably we should rename this trait and add new methods like `generate_instance_assignment` and `generate_witness_assignment`

Alternatively, we could replace the trait with three closures that do these tasks, as the TODO suggests.

snark/src/r1cs.rs
Pratyush: Instead of having these segregated APIs, another idea would be to add a `RelationProvider` trait that is a generalization of `RelationProvider`.

```rust
pub trait RelationProvider<R: Relation> {
	// These are all optional because they might not exist always.
	// For example, in indexing phase, there is no assignment,
	// while during proving there may be no index.
	fn index(&self) -> Option<R::Index>; // Maybe this should be `Rc<RefCell<R::Index>>`? to avoid copying.
	fn instance(&self) -> Option<R::Instance>;
	fn witness(&self) -> Option<R::Witness>;
}

pub trait SNARK<R: Relation> {
	fn prove<RP: RelationProvider<R>, ...>(pk, rp: RP)
	// similar for other methods.
}

impl<F: Field> RelationProvider<R1CS<F>> for ConstraintSynthesizer<F> {
// do stuff.
}
```

relations/src/r1cs/constraint_system.rs
weikengchen: Looks good.

snark/src/r1cs.rs
```diff
@@ -0,0 +1,70 @@
...
+        proof: &Self::Proof,
+    ) -> Result<bool, Self::Error>;
+}
+
+/// A SNARK with (only) circuit-specific setup.
+pub trait CircuitSpecificSetupSNARKForR1CS<F: Field>: CircuitSpecificSetupSNARK<R1CS<F>> + SNARK<R1CS<F>> {
+    /// The setup algorithm for circuit-specific SNARKs.
+    fn setup<C: ConstraintSynthesizer<F>, Rng: RngCore + CryptoRng>(
```
weikengchen: Should we have a more specific name like `circuit_specific_setup`?

Pratyush: Yeah that makes sense, I’ll do that

snark/src/lib.rs
```diff
@@ -38,34 +41,32 @@ pub trait SNARK<R: NPRelation> {
     /// Generates a proof of satisfaction of the arithmetic circuit C (specified
     /// as R1CS constraints).
     fn prove<Rng: RngCore + CryptoRng>(
-        circuit_pk: &Self::ProvingKey,
+        pk: &Self::ProvingKey,
```
ValarDragon, line 44: I feel like we should be more explicit with the naming here, e..g `proving_key`, since one could confuse it with a public_key / private_key

snark/src/r1cs.rs
```diff
@@ -1,70 +1,48 @@
...
 
-/// The basic functionality for a SNARK.
+/// A [`SNARKForR1CS`] is a [`SNARK`] for the [`R1CS`] relation.
 pub trait SNARKForR1CS<F: Field>: SNARK<R1CS<F>> {
-    /// Generates a proof of satisfaction of the arithmetic circuit C (specified
-    /// as R1CS constraints).
-    fn prove<C: ConstraintSynthesizer<F>, Rng: RngCore + CryptoRng>(
-        circuit_pk: &Self::ProvingKey,
-        witness_generator: C,
+    /// Generate inputs for the SNARK indexer from [`cs`].
```
ValarDragon, line 12:
```suggestion
    /// Generate the matrices for the SNARK indexer from [`cs`].
```

snark/src/r1cs.rs
```diff
@@ -1,70 +1,48 @@
...
+    fn indexer_inputs<CS: ConstraintSynthesizer<F>>(cs: &CS) -> ConstraintMatrices<F>;
+
+    /// Generate inputs for the SNARK prover from [`cs`].
+    fn prover_inputs<CS: ConstraintSynthesizer<F>>(cs: &CS) -> (Instance<F>, Witness<F>);
+
+    /// Generate inputs for the SNARK verifier from [`cs`].
```
ValarDragon, line 26:
```suggestion
    /// Generate inputs for the SNARK verifier from [`cs`].
    /// Assumes the verifier has a version of the matrices from elsewhere.
    /// (Either preprocessed, or a direct copy of the matrices)
```

snark/src/r1cs.rs
```diff
@@ -1,70 +1,48 @@
...
+    fn verifier_inputs<CS: ConstraintSynthesizer<F>>(cs: &CS) -> Instance<F>;
+
+    /// Generates a proof of satisfaction of the constraint system induced by [`cs`].
```
ValarDragon:
```suggestion
    /// Generates a proof of satisfaction of the constraint system with inputs and witness induced by [`cs`].
```

ValarDragon (approved):
LGTM! This seems like a significant improvement to the interfaces here

ValarDragon:
Accidentally just reviewed the last few commits lol

relations/src/lib.rs
```diff
@@ -16,3 +16,25 @@
 extern crate ark_std;
 
 pub mod r1cs;
+
+/// An *indexed relation* is a set of triples of the form `(index, instance, witness)`.
```
ValarDragon, line 20:
```suggestion
/// An *indexed relation* is a set of triples of the form `(index, instance, witness)`.
/// We often want to test if `(index, instance)` is in the relation.
```

relations/src/r1cs/mod.rs
```diff
@@ -1,9 +1,125 @@
...
+//! Core interfaces for working with Rank-1 Constraint Systems (R1CS), which is
+//! an indexed NP relation.
 
-use ark_std::vec::Vec;
+use crate::{NPRelation, Relation};
+use ark_std::{cmp::Ordering, marker::PhantomData, vec::Vec};
 
-/// A result type specialized to `SynthesisError`.
-pub type Result<T> = core::result::Result<T, SynthesisError>;
+/// An *indexed NP relation* is a relation with an efficient membership check.
```
ValarDragon:
```suggestion
/// R1CS is an *indexed NP relation*. 
/// An index consists of three matrices (A, B, C), 
/// while the instance *x* and witness *w* are vectors of field elements 
/// such that, for z := (x||w), Az ○ Bz = Cz
```

relations/src/r1cs/mod.rs
```diff
@@ -1,9 +1,125 @@
...
+    pub num_constraints: usize,
+    /// The number of non_zero entries in the A matrix.
+    pub a_num_non_zero: usize,
+    /// The number of non_zero entries in the B matrix.
+    pub b_num_non_zero: usize,
+    /// The number of non_zero entries in the C matrix.
+    pub c_num_non_zero: usize,
+
+    /// The A constraint matrix.
+    /// `self.mode == SynthesisMode::Prove { construct_matrices = false }`.
```
ValarDragon: Whats going on with this line?

relations/src/r1cs/mod.rs
```diff
@@ -1,9 +1,125 @@
...
+    pub a: Matrix<F>,
+    /// The B constraint matrix.
+    pub b: Matrix<F>,
+    /// The C constraint matrix.
+    pub c: Matrix<F>,
+}
+
+/// An R1CS instance consists of variable assignments to the instance variables.
+/// The first variable must be assigned a value of `F::one()`.
```
ValarDragon: I wonder if we should do the converse, and make the second variable onwards what goes into the struct?

relations/src/r1cs/mod.rs
```diff
@@ -1,9 +1,125 @@
...
+        // The number of witness variables does not match.
+        if witness.0.len() != index.num_witness_variables {
+            return false;
+        }
+        // The first instance variable must be 1.
+        if instance.0[0] != F::one() {
+            return false;
+        }
+
+        cfg_iter!(&index.a)
```
ValarDragon, line 79:
```suggestion
        // Let z = instance || witness. Check that Az ○ Bz = Cz, where ○ is the Hadamard product.
        // The Hadamard product of two vectors is the product of them entry-wise, so
        // [ 2 ]   [ 3 ]   [ 6 ]
        // [ 3 ] ○ [ 6 ] = [ 18 ]
        // [ 4 ]   [ 7 ]   [ 28 ]
        cfg_iter!(&index.a)
```

snark/src/lib.rs
Pratyush: doesn't the type clarify that?

snark/src/r1cs.rs
ValarDragon: Or perhaps we should keep this line, and a second line in the comments for each of these, specifying what the inputs are?

relations/src/r1cs/mod.rs
Pratyush:
```suggestion
```

snark/src/r1cs.rs
Pratyush: I like that latter idea.

relations/src/r1cs/mod.rs
Pratyush: what do other sources of R1CS do, like zkinterface?

snark/src/lib.rs
ValarDragon: You're right, forgot it appeared in the docs

snark/src/r1cs.rs
Pratyush: Done.

weikengchen (approved):

Pratyush:
Updated the PR further, I think it's now in the final form and is ready for merging as is (modulo the CHANGELOG)

snark/src/lib.rs
```diff
@@ -101,30 +93,53 @@ pub enum UniversalSetupIndexError<Bound, E> {
 
 /// A SNARK with universal setup. That is, a SNARK where the trusted setup is
 /// circuit-independent.
-pub trait UniversalSetupSNARK<F: PrimeField>: SNARK<F> {
+pub trait UniversalSetupSNARK<R: NPRelation>: SNARK<R> {
     /// Specifies how to bound the size of public parameters required to
     /// generate the index proving and verification keys for a given
     /// circuit.
```
ValarDragon, line 100: Lets add an example of what an index bound would be for a particular universal setup? (namely a degree bound for KZG)

Pratyush: Done.

jon-chuang:
So the `Relation` trait is pretty generic, and allows for a generic `Snark` interface. Which is nice. 

So I'm wondering if it's a good time to try to do, or at least plan for, a draft migration of the [halo2](https://github.com/zcash/halo2) code, although @daira seemed to be saying its not exactly ready for that. But from my perspective, it maybe would help put an extra pair of eyes on zcash's code and help arkworks advance to being yet more featured and flexible. 

I'll post an issue soon regarding an approach, also outlining potential friction areas and incompatibilities, and ideas for how to solve them.

I'll also try to see if there are any tweaks that could be made regarding this PR.

---

I'm also wondering if relatedly, a common "plonk description language" ought to be established to make at the very least, relations, which can consist of multiple sub-relations, compatible between various libraries. The description language can also contain meta data, such as subcircuit names and component names, or not.

If R1CS is a general-purpose processor, plonk/plookup is a sort of FPGA/ASIC, with reuseable components, lookup tables, wiring, fan-in/fan-out and locality constraints, custom gates. The glue that holds it all together is the wiring (permutation argument). There are also subtle tradeoffs between the maximum relation multiplicative depth and proof size/proving time.

Just like the differences between programming a CPU and programming an FPGA are huge, there is a need to manage, and hide, a lot more complexity when it comes to Plonk/Plookup.

relations/src/lib.rs
```diff
@@ -16,3 +16,25 @@
...
+pub trait Relation {
+    /// The index is a "large" but static part of the triple. Examples include
+    /// the circuit in the circuit satisfiability relation, and constraint
+    /// matrices in the R1CS relation.
+    type Index: Eq;
+    /// The instance is a "small" part of the triple. Like the index, it is publicly known.
+    type Instance: Eq;
+    /// The instance is a "large" but private part of the triple.
```
Will-Lin4: instance -> witness?

relations/src/r1cs/constraint_system.rs
```diff
@@ -85,14 +96,17 @@ pub enum SynthesisMode {
     /// Indicate to the `ConstraintSystem` that it should only generate
     /// constraint matrices and not populate the variable assignments.
     Setup,
-    /// Indicate to the `ConstraintSystem` that it populate the variable
+    /// Indicate to the `ConstraintSystem` that it should populate the witnessvariable
```
Will-Lin4: witnessvariable -> witness variable?

weikengchen:
When should we merge this one (and plan for the next/next release)? And how many code changes to the other repos should we expect?

Pratyush:
I still have to make corresponding PRs for Groth16 and Marlin; let's push merging this until after 0.3 (so in 0.4).

PR #362, opened 2022-08-28
Prepare ark-snark for the new updates in ark-algebra

weikengchen:
## Description

This draft PR has several changes: (1) add patching, (2) change ark-test-curves feature to the curve one temporarily before a corresponding PR in ark-algebra is merged, and (3) fix a confusing "type annotations needed".

Since (2) is in a pending PR, it would be best not to merge this PR before that one is done, specifically giving that we are working on the `snark` repo. 

https://github.com/arkworks-rs/algebra/pull/461

---

write a little note why.

N/A:
- [ ] Wrote unit tests
- [ ] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`


relations/src/r1cs/constraint_system.rs
```diff
@@ -340,7 +341,11 @@ impl<F: Field> ConstraintSystem<F> {
                     let lc = transformed_lc_map
                         .get(&lc_index)
                         .expect("should be inlined");
-                    transformed_lc.extend((lc * coeff).0.into_iter());
```
Pratyush: Why did this change?

Pratyush:
This LGTM modulo the comment

relations/src/r1cs/constraint_system.rs
findora-crypto: In my local attempt it appears that "<&LinearCombination<F> as Mul<F>>::mul" is needed, otherwise the compiler is complaining. Actually, let me try again to see if this is still the case.

weikengchen: A smaller change is made, so we continue to use `transformed_lc.extend((lc * coeff).0.into_iter());`

with the only change in signaling to the compiler the following:
```diff
- let mut transformed_lc_map = BTreeMap::new();
+ let mut transformed_lc_map = BTreeMap::<_, LinearCombination<F>>::new();
```

weikengchen: Just found out "```diff"

weikengchen: An approval will be needed.

snark/src/lib.rs
```diff
@@ -23,7 +23,7 @@ pub trait SNARK<F: PrimeField> {
 
     /// The information required by the verifier to check a proof for a specific
     /// circuit *C*.
-    type VerifyingKey: Clone + ToBytes;
+    type VerifyingKey: Clone;
```
Pratyush: Should we replace this with a `CanonicalSerialize + CanonicalDeserialize` bound?

Pratyush: (Same for `ProvingKey`, `ProcessedVerifyingKey`, and `Proof`)

weikengchen:
^ The change above (using CanonicalSerialize + CanonicalDeserialize) is done. But, it depends on https://github.com/arkworks-rs/algebra/pull/476

weikengchen:
Closed --- needs to rebuild the PR.

PR #380, opened 2024-11-01
`arithmetic-circuits` module

Cesar199999:
## Description

This PR adds a new crate, `arithmetic-circuits`, which contains the structures `ArithmeticCircuit` and `Expression`, as well as methods for compiling R1CS constraint systems (`ark_relations::r1cs::ConstraintSystem`) into arithmetic circuits. 
An `ArithmeticCircuit` is a directed acyclic graph where nodes are either variables, constants, or gates for addition and multiplication over a prime field `F`. The circuit is internally represented as a list of nodes, each being a variant of the enum `Node`: `Variable`, `Constant`, `Add`, and `Mul`. Gates have a left and right input, which are indices for the respective nodes in the array. In order to directly construct an arithmetic circuit, the user must instantiate an empty `ArithmeticCircuit` struct and mutate it via its public methods (e.g. `new_variable`, `add`, `mul`), each of which returns the index of the new node in the array.  For example, an  arithmetic circuit expressing the computation `2 * x + y` can be constructed as follows:

```rust
let mut circuit = ArithmeticCircuit::new();
let two = circuit.constant(F::from(2));
let x = circuit.new_variable();
let y = circuit.new_variable();
let two_x = circuit.mul(two, x);
let result = circuit.add(two_x, y);
```

Since the construction of an arithmetic circuit can be slightly verbose, the crate also provides the struct `Expression`, which allows for a more user-friendly way to construct arithmetic circuits. Manually adding nodes to the circuit is not required: expressions can be combined using the usual arithmetic operators, and then converted to a circuit using the `to_arithmetic_circuit` method. For reference, the previous circuit can be expressed as follows: 
```rust
let x = Expression::variable("x");
let y = Expression::variable("y");
let circuit = (2 * x + y).to_arithmetic_circuit();
```

Note that the variables must be labelled uniquely.

Lastly, the crate provides methods for compiling R1CS constraint systems from circom generated `.r1cs` and `.wasm` files into arithmetic circuits. 

Discussion NP-Eng/ligero#25


mmagician:
@Pratyush, as per the discussion in https://github.com/NP-Eng/ligero/issues/25#issuecomment-2346989554, this PR adds the arithmetic circuit functionality, ready for review when you get some cycles

Pratyush:
Do you think this would be a better fit in ark-relations? Since r1cs (another satisfiability relation) lives in there?

Cesar199999:
@Pratyush Sure, that makes sense

PR #410, opened 2025-08-09
Fix potential UB in `Variable::kind()` method

anim001k:
### Description
Replaces `unsafe { unreachable_unchecked() }` with a safe fallback in `Variable::kind()` to prevent undefined behavior when encountering invalid variable tags.

### Changes
-  Added `debug_assert!(tag <= 4)` to catch invalid tags in debug builds
-  Replaced UB-prone `unreachable_unchecked()` with safe default return
-  Preserved `const fn` signature and performance characteristics


Pratyush:
please don't make AI-driven low-effort PRs.

PR #415, opened 2025-09-21
fix: LinearCombination::get_var_loc linear-search to return Ok on equality

radik878:
- The small-vector branch of get_var_loc always returned Err, even when the searched Variable existed, causing AddAssign<(F, Variable)> to insert duplicates instead of summing coefficients.
- This change aligns the linear-search behavior with binary_search semantics: Ok(i) when equal, Err(i) at the first greater element, Err(len) if not found.
- Ensures LC invariants during incremental construction for short combinations and prevents duplicate entries that inflate LC weight.

### DIFF
--- a/relations/src/utils/linear_combination.rs
+++ b/relations/src/utils/linear_combination.rs
@@ -599,3 +599,31 @@ impl<F: Field> Sub<(F, LinearCombination<F>)> for LinearCombination<F> {
         self + (-coeff, other)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::LinearCombination;
+    use crate::utils::variable::Variable;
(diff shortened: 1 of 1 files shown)


Pratyush:
Can you please add a test which demonstrates that the old code was broken? Ideally by splitting the change into two commits: the first with the test, and the second with the fix?

radik878:
> Can you please add a test which demonstrates that the old code was broken? Ideally by splitting the change into two commits: the first with the test, and the second with the fix?

Added a test case, but wanted to ask you, is this ok if the first commit with the fix, and second with test? or it neccessarily shoud be first with test and second with fix?

Pratyush:
Sure.

PR #416, opened 2025-09-22
Replace unreachable_unchecked in Variable::kind to avoid UB

viktorking7:
### Description
- **Summary**: Replace `core::hint::unreachable_unchecked()` with a safe `unreachable!()` in `relations/src/utils/variable.rs` (`Variable::kind`).
- **Rationale**: Prevents potential undefined behavior if an invalid variable tag is ever observed (e.g., due to data corruption or invariant violations).
- **Change**: 
  - `_ => unsafe { core::hint::unreachable_unchecked() }` → `_ => unreachable!("invalid variable tag: {}", self.tag())`
- **Impact**: No runtime overhead in the common path; in anomalous cases, it fails fast with a clear error message instead of causing UB.


Pratyush:
please don't make AI PRs.

